PR #8, opened 2020-12-14
Fix python backend for 2019

### LINKED ISSUE #6: Fix rounding in Python backend
The rounding function in the Python backend of Mlang :

https://github.com/MLanguage/mlang/blob/a18b81e0f5b0e2ab61a460d03a2df2167993ac88/src/mlang/backend_compilers/bir_to_python.ml#L84-L86

Does not match the new (and correct) definition of the rounding function in the Mlang interpreter, which used casting back and forth from floats to integers:

https://github.com/MLanguage/mlang/blob/a18b81e0f5b0e2ab61a460d03a2df2167993ac88/src/mlang/backend_ir/bir_interpreter.ml#L45-L46

The Python function should be updated to reflect this behavior.

rmonat:
Closes #6


denismerigoux (approved):
Great, thanks !

PR #10, opened 2020-12-16
Remove references to nonexistant F* formalization

jemsab:
F* formalization was previously removed from project, this commits removes references to that file from the main README.


denismerigoux:
Indeed, nice catch, thanks!

PR #11, opened 2020-12-22
Fixing error handling

denismerigoux:
It turns out fuzzing the 2019 version of the published code revealed another error in Mlang. Indeed, there was a problem concerning the way Mlang handles errors.

## Semantics of logical operators

First, the logical operator `OR` actually has an additive, neutral semantics with respect to `undefined`. `OR` follows addition, and differs from `AND` or multiplication. 

## Fixing errors

A lot of the error handling is conditioned to the fact that the `APPLI_OCEAN` variable is set to 0. This variable is set to 0 in some part of the INTER code of the DGFiP, so we have to add that part that was missing in our M++ files.

Second, the behavior of DGFiP is that only "anomalies" errors cause the computation to stop. "discordance" and "information" errors allow the computation to carry through until the end. We replicate this behavior by just discarding the "discordance" and "information" verification conditions in the M frontend -> MIR translation.

## Removing faulty fuzzing tests

Because of these changes in the error handling that don't change anything in the way Mlang passes the DGFiP private tests, fuzzer-generated tests expect the wrong behavior for errors. As such, a large portion of them are invalidated and deleted from this PR. Hence, they should be re-generated.

**Missing before merging:** re-generate fuzzing tests for 2018 and 2019


denismerigoux:
The backend tests are not working anymore because partial evaluation on `OR` has to be fixed here:
https://github.com/MLanguage/mlang/blob/2057e3dbf46cb97c94687f313d000f797430b872/src/mlang/optimizing_ir/partial_evaluation.ml#L257-L269

rmonat (approved):
Everything seems good to me, I tested locally. We just need the updated tests. Then maybe we can move the CI to the 2019 version?

PR #14, opened 2021-01-12
Java backend

jemsab:


denismerigoux:
Thanks James ! I did a merge to clean the `git diff` view. A quick suggestion before I do a full review : a long time ago, I did a Java backend for Mlang. Here is the file : https://github.com/MLanguage/mlang/blob/99df4f8694b17651673bbf101c841f5837e2b381/src/mlang/backends/java/mvg_to_java.ml. The code in there will likely not work any more since the compiler has changed a lot but you can take inspiration from it I guess.

denismerigoux:
Another comment I had while reviewing your changes: I would suggest you create `examples/java/MValue.class` on the model of `examples/c/m_value.{c,h}`, an external class file containing all the operations for the values manipulated by the main program. You could move there the content of your `m_operation_class` variable.

I know the Python backend re-generates the `Undefined` class all the times but in restrospect this should also be changed to adopt this model of external library in the target language.

denismerigoux:
Yea the 64kb limit for methods code in Java is a pain, you have to artifically cut your generated code into little pieces...

denismerigoux:
So in the old backend the way I solved it I think was to move the definition of each M variable into its own method, and then creating a bit last method to chain all the small method calls on the variable container.

jemsab:
Indeed, I'm also using that solution but it's still too big. 
I'll have to keep splitting. 
The code I've written so far is a bit of a mess, I'll have to refactor much of it. A rewrite splitting everything up into lists that can be placed anywhere may be necessary.

denismerigoux:
Is it the 64kB method limit you're hitting or the parser limit with the number of opened parenthesis ? For the latter you can take inspiration from 

https://github.com/MLanguage/mlang/blob/f5603966acc4272615dc569132b043902cc4d66b/src/mlang/backend_compilers/bir_to_python.ml#L163-L190

to reduce the number of parenthesis needed.

jemsab:
I'm getting "code too large" on the main calculation method, as I understand it that means the compiler is hitting the 64KB limit.

denismerigoux:
Hum in that case you should do a two-level hierarchy : first you generate one method per ~1000 M variable assignations, then you create a final method calling all the sub-methods in order, each one representing ~1000 M variables. It's dumb but Java is quite adverse to doing what we're doing :)

denismerigoux:
I see you're refactoring `list` to  `Queue.t`, I suspect you're having performance problems in the backend. I don't think this is due to `list` vs `Queue.t`, but rather to the fact that you are storing intermediate program nodes as `string` before flushing them to the file where you're writing the Java output. If you want an efficient Java backend, then you should pass the control flow in OCaml such that you don't store what you want to print in intermediate OCaml data structures, but rather `Format.fprintf` it right away.

jemsab:
You're exactly right, as you noted I'm storing every node as a `string` which obviously uses a ton of RAM, especially without the optimisation flag off.
It did seem like a simpler solution to get something minimal working as well as something more understandable. 

I hadn't thought of a way to write in two different places in a file in OCaml, but after having read your comment I've found out about `seek_out`. 

I guess I just need to track the position of the last method call as well as the last method definition so as to be able to jump between the two. I will also need to track the number of lines / assignations so I know when to jump. Thanks for the guidance!

denismerigoux:
I didn't even know about `seek_out`, I was more thinking about delaying the stuff that you want to write by passing it a a closure (let f = `fun _ -> ... in ...`) to later call it (`f ()`). We can talk more about it at the end of the week :)

jemsab:
Ah I see! After basic testing seek_out isn't going to help, I can't seem to append to a file mid-file (which unfortunately makes sense). 
I'll think about your closure idea to try and understand the implementation before Wednesday so we can discuss it further.

examples/java/src/com/mlang/MValue.java
```diff
@@ -100,21 +100,49 @@ public static OptionalDouble mOr(OptionalDouble value1, OptionalDouble value2) {
...
-    return binopCondition((firstValue, secondValue) -> {
-      return OptionalDouble.of(firstValue.getAsDouble() + secondValue.getAsDouble());
-    }, value1, value2);
+
+    if(value1.isEmpty() && value2.isEmpty()) {
+      return OptionalDouble.empty();
+    }
+
+    double localValue1 = 0.;
+    double localValue2 = 0.;
```
denismerigoux, line 109: Wouahou that's painful. Maybe it's better to define a custom `MValue` class with a value field and an undefined field, that follows more closely what is done in `m_value.c` ?

jemsab: I can certainly transform `OptionalDouble` into a specific class but as I understand that would only partially solve the problem as you can't perform logical operations ( `||` or `&&` ) on integers in Java

denismerigoux: >  that would only partially solve the problem as you can't perform logical operations ( || or && ) on integers in Java

What is the problem with logical operations ? The way logical operations are defined in M follows what happen with bitwise operations (`|`, `&` in C), treating 0 as false and any non-zero as true.

jemsab: `x.value && y.value` does not work in Java, so you need a way to check if values are equal to 0.

For example, two integers toto and tata give this error message : 
```
bad operand types for binary operator '&&'
|    first type:  int
|    second type: int
|  tata && toto;
|  ^----------^
```

denismerigoux: Yep so you can do `x.value & y.value` to go the right result non ?

jemsab: In the m_value.c file I have it is && and ||, do I have a wrong version somehow ? 

Also using single &  doesn't seem to wotk either unfortunately : 

Values are 0 : 

```
jshell> tata = 0;
tata ==> 0

jshell> toto = 0;
toto ==> 0

jshell> tata & toto;
$17 ==> 0

```

Values are <> 0 :
```
jshell> toto = 10;
toto ==> 10

jshell> tata = 20;
tata ==> 20

jshell> tata & toto;
$20 ==> 0
```

denismerigoux: Ok here's the spec of `&&` and `||` in C: https://docs.microsoft.com/en-us/cpp/c-language/c-logical-operators?view=msvc-160. They match the semantics of `&` and `|` if and only if all the values manipulated are either 0 or 1. So it seems you would need a custom `m_and` and `m_or` function in Java to replicate the C semantics of `&&` and `||`, because we can't assume that M values that act as booleans are constrainted to either 0 or 1.

jemsab: I'm not sure I follow how my current (if inelegant) implementation should be changed to match those custom semantics

denismerigoux: You current implementation of `mAnd` and `mOr` are perfectly fine. My original comment concerned `mAdd` and `mSubstract` which have to use local variables.

jemsab: Oh of course, I beg your pardon, I got my paint brushes mixed up (to use a Frenchism), indeed a mvalue type similar to C would certainly give cleaner code. I'll put it down as a todo.

I've got the basic_case calculator working in Java so I think we're getting close to having a working java version, just few more details to iron out in the implementation and with the test harness.

examples/java/Makefile
```diff
@@ -0,0 +1,43 @@
...
+	$(MLANG) \
+		--backend java --output src/com/mlang/Ir_tests_2020.java \
+		--function_spec $^ \
+		$(SOURCE_FILES)
+
+target/com/mlang/Ir_%.class:  src/com/mlang/Ir_%.java
+	javac -target 1.7 -source 1.7 -J-Xss10m -d target -cp src src/com/mlang/*.java 
+
+backend_tests/target/TestHarness.class: target/com/mlang/Ir_tests_2020.class
+	javac -J-Xss10m -cp "backend_tests/src/:target" -d backend_tests/target backend_tests/src/com/mlang/TestHarness.java
```
denismerigoux, line 40: Maybe document what the `-J-Xss10m` flags are doing?

m_specs/tests_2020.m_spec
```diff
@@ -0,0 +1,7 @@
...RILMNP3, BTRILMNP4, BTRILMNP5, BTRNI, BTRRCEL1, BTRRCEL2, BTRRCEL3, BTRRCEL4, BTSANCTION, BTXYZ1, BTXYZ2, CHC, CHI, CIC, CII, CIT, CJC, CJG, CJI, CKC, CKI, CLC, CLI, CMI, CNC, CNI, CNS, COC, COI, CPS, CPT, CPU, CQC, CQI, CRC, CRF, CRI, CSC, CSI, CSN, DAJ, DPS, DPT, DPU, EBJ, HBI, HBJ, HBK, HBM, HBN, HBO, HBP, HBR, HBS, HBT, HBW, HBX, HBY, HBZ, HCB, HCC, HCD, HCE, HCG, HCI, HCJ, HCK, HCM, HCN, HCO, HCP, HCR, HCS, HCT, HCU, HCW, HDJ, HDK, HDM, HDN, HDO, HDP, HDR, HDS, HDT, HDU, HDW, HEN, HEO, HEP, HER, HES, HET, HEU, HEW, HFN, HFO, HFP, HFR, HFS, HFT, HFU, HFW, HGS, HGT, HGU, HGV, HGW, HHC, HJA, HJC, HKC, HOB, HOC, HOI, HOK, HOP, HOQ, HOW, HOY, HQA, HQB, HQC, HQD, HQL, HQM, HQT, HUA, HUF, HUG, HUH, HUI, HUL, HUM, HUN, HUO, HUP, HUQ, HUR, HUS, HUT, HUU, HVA, HVB, HVC, HVD, HVE, HVF, HVG, HVH, HVI, HVJ, HXF, HXG, HXH, HXI, HXK, HXL, HXM, HXN, HXO, HXP, HXQ, HXR, HXS, HXT, HXU, HYA, HYB, HYC, HYD, HYE, INDIDF, RAJ, RAS, RAW, RBA, RBK, RBO, RBP, RBS, RBT, RCF, RCH, RCK, RCP, RCS, RDO, RDP, RDW, REGCO, REM, REP, RFM, RTR, RVA, RYY, SAA, SAB, SAC, SAD, SAE, SAF, SAG, SAH, SAI, SAJ, SAK, SAL, SAO, SAP, SAQ, SAR, SAS, SAT, SAU, SAV, SAW, SAX, SBA, SCA, SDA, SDE, SDP, SDQ, ZRA, ZRB, ZRS, ZRU;
```
denismerigoux, line 1: Good catch, we hadn't checked this into GitHub yet.

src/mlang/backend_compilers/bir_to_c.ml
```diff
@@ -259,31 +259,6 @@ let generate_main_function_signature (oc : Format.formatter) (add_semicolon : bo
...
     (if add_semicolon then ";" else "")
 
-let get_variables_indexes (p : Bir.program) (function_spec : Bir_interface.bir_function) :
-    int Mir.VariableMap.t * int =
-  let input_vars = List.map fst (VariableMap.bindings function_spec.func_variable_inputs) in
-  let assigned_variables =
-    List.map snd (Mir.VariableDict.bindings (Bir.get_assigned_variables p))
-  in
-  let output_vars = List.map fst (VariableMap.bindings function_spec.func_outputs) in
-  let all_relevant_variables =
```
denismerigoux, line 269: Good refactoring move

src/mlang/backend_compilers/bir_to_c.ml
```diff
@@ -519,7 +486,8 @@ let generate_c_program (program : Bir.program) (function_spec : Bir_interface.bi
       (Format.asprintf "Output file should have a .c extension (currently %s)" filename);
   let header_filename = Filename.remove_extension filename ^ ".h" in
   let _oc = open_out header_filename in
-  let var_indexes, var_table_size = get_variables_indexes program function_spec in
+  let var_indexes, var_table_size = Bir_interface.get_variables_indexes program function_spec in
+  Printf.printf "var_table_size %d\n" var_table_size;
```
denismerigoux, line 490: This can be removed I guess

src/mlang/backend_compilers/bir_to_java.ml
```diff
@@ -0,0 +1,426 @@
...
+        (fun fmt ->
+          IndexMap.iter (fun i v ->
+              let sv, defs = generate_java_expr v var_indexes in
+              Format.fprintf fmt "%acalculationVariables[%d /* %a */] = %s;@,"
+                (format_local_vars_defs var_indexes)
+                defs
+                (get_var_pos var var_indexes |> ( + ) i)
+                format_var_name var sv))
+        es
+      (* Format.fprintf oc "mCalculation.getTableVariables().put(\"%a\",Arrays.asList(%a));@,"
```
denismerigoux, line 195: Debugging comment that can be removed ?

src/mlang/backend_compilers/bir_to_java.ml
```diff
@@ -0,0 +1,426 @@
...
+  let count = List.fold_left (fun count assign -> format_input_var assign count) 0 input_vars in
+  Format.fprintf oc
+    "}@,\
+     static void loadInputVariables(Map<String, MValue> inputVariables, MValue[] \
+     calculationVariables) {@,\
+     %a@,\
+     }"
+    (print_load_input 0) (count / split_threshold)
+
+let sanitize_str (s, p) =
```
denismerigoux, line 242: Maybe a little doc that says that the DGFiP source files are encoded with ISO something and that they might contain characters that Java doesn't like?

src/mlang/backend_compilers/bir_to_java.mli
```diff
@@ -0,0 +1 @@
+val generate_java_program : Bir.program -> Bir_interface.bir_function -> string -> unit
```
denismerigoux, line 1: The `.mli` should also contain the copyright header.

src/mlang/backend_ir/bir.ml
```diff
@@ -34,6 +34,39 @@ type program = {
   outputs : unit Mir.VariableMap.t;
 }
 
+let squish_statements (program : program) (threshold : int) (rule_suffix : string) =
```
denismerigoux, line 37: This should be documented from a high-level perspective.

src/mlang/backend_ir/bir.ml
```diff
@@ -34,6 +34,39 @@ type program = {
...
+  let rule_from_stmts stmts =
+    let id = Mir.fresh_rule_id () in
+    { rule_id = id; rule_name = rule_suffix ^ string_of_int id; rule_stmts = List.rev stmts }
+  in
+  let rec browse_bir old_stmts new_stmts curr_stmts rules =
```
denismerigoux, line 42: There should be type annotations here.

src/mlang/backend_ir/bir.ml
```diff
@@ -34,6 +34,39 @@ type program = {
...
+        let rules, curr_stmts =
+          match Pos.unmark hd with
+          | SConditional (expr, t, f) ->
+              let t_rules, t_curr_list = browse_bir t [] [] rules in
+              let f_rules, f_curr_list = browse_bir f [] [] t_rules in
+              let cond = give_pos (SConditional (expr, t_curr_list, f_curr_list)) in
+              (f_rules, cond :: curr_stmts)
+          | _ -> (rules, hd :: curr_stmts)
+        in
+        (* TODO: The following line works because of a low threshold. In order to make this robust,
```
denismerigoux, line 56: `List.length (List.flatten curr_stmts)` ?

src/mlang/backend_ir/bir.mli
```diff
@@ -34,6 +34,8 @@ type program = {
   outputs : unit Mir.VariableMap.t;
 }
 
+val squish_statements : program -> int -> string -> program
```
denismerigoux, line 37: High-level doc should go here.

denismerigoux (changes requested):
Bravo James! This is a major achievement that provides major new functionality to Mlang. The ability to target Java will surely be of great interest to the DGFiP. Now, they simply can embark the 6,6 MB compiled "calculette" into their applications to get the "primitif" computation for free from the M/M++ codebase. This is a real success story, I hope it gets publicized internally.

I requested some nitpicks changes in my comments, but it's mostly documentation, you've already done the hard work. Also there's a `README.md` missing in the `example/java` directory to indicate how your Java backend is organized and how to use it as an external user. You can get some inspiration from `'example/python/README.md`, and update the top-level `README.md` to indicate that we now support Java.

examples/java/Makefile
jemsab: These options determine maximum stack size usable by the JVM. Now we've split the rules and outsourced input handling to another class, these overrides no longer seem necessary, at least to run tests. I'm removing them from the Makefile and it will be up to the client user of Mlang to set different defaults should they so wish.

src/mlang/backend_ir/bir.ml
jemsab: Unfortunately `curr_stmts` isn't a `'a list list` . Testing in utop and directly in the code doesn't seem to want to work. I think the only choice here is to recursively descend into the list, counting the lists inside `SConditional`s as we go.

Maybe I'm missing something?

denismerigoux: I found the solution with 10cbf45, you can take a look at it :)

src/mlang/backend_compilers/bir_to_java.ml
jemsab: Refactored into strings module for python and Java backends and given proper documentation

examples/java/README.md
```diff
@@ -0,0 +1,79 @@
...
+    make src/com/mlang/Ir_<name_of_the_m_spec_file>.java
+
+`make` will re-generate this file at each modification of the `.m_spec` file,
+otherwise use `-B` to force re-generation.
+
+If you have an error about unknown variables, then it probably means the
+example you're trying to build depends on a different year of the tax code.
+To get it working, invoke the command by setting the year to another year:
+
+    YEAR=<year> make ir_<name_of_the_m_spec_file>.py
```
denismerigoux, line 70: Missing a `.py` :)

denismerigoux (approved):
Looks very good to me! I'll wait for the typo correction and merge :)

PR #15, opened 2021-02-13
NaN optimization

dmentre:
Hello,

Here is a first PR to add an "--c-nan-opt" option to mlang that produces code using a single float with NaN value for undefined M value. From my test, one gains 23 to 25%.

I had to duplicate code of test_harness.c and perf_harness.c because the code was too specific to struct m_value representation. Maybe there is a way to have a common API.

Let me know what you think of it.

Best regards,
david


denismerigoux:
Hi David ! First of all thanks for your contribution, which is a follow up to our email discussions :)

I think that your `n_nan_value.{c,h}` is fine, as well as your modifications to the OCaml backend of Mlang. However, this duplication of the harnesses code is indeed unnecessary. 

One could avoid the duplication by including at the top of `perf_harness.c` and `test_harness.c` some `typedef` enclosed within `#ifdef` macros that would expand to either `m_value.h` definitions of `m_nan_value.h` definitions dependending on some compile-time flag `USE_NAN_OPTIM`. This compile time flag could be set directly in the `ir_*.h` file generated by Mlang, depending on whether the `--c-nan-opt` flag has been passed to Mlang.

dmentre:
Hello Denis,

Le 14/02/2021 à 12:09, Denis Merigoux a écrit :
>
> However, this duplication of the harnesses code is indeed unnecessary.
>
> One could avoid the duplication by including at the top of 
> |perf_harness.c| and |test_harness.c| some |typedef| enclosed within 
> |#ifdef| macros that would expand to either |m_value.h| definitions of 
> |m_nan_value.h| definitions dependending on some compile-time flag 
> |USE_NAN_OPTIM|. This compile time flag could be set directly in the 
> |ir_*.h| file generated by Mlang, depending on whether the 
> |--c-nan-opt| flag has been passed to Mlang.
>
I pushed such modification in my NaN branch which should be available 
within the pull request.

Depending on |USE_NAN_OPTIM| defined or not, both perf_harness.c and 
test_harness.c include ir_tests.nan.h or ir_tests.h and make relevant 
definitions for following code. I removed no longer needed 
perf_nan_harness.c and test_nan_harness.c. I updated the Makefile 
accordingly.

Let me know what you think of it.

Best regards,
david

examples/c/m_nan_value.c
```diff
@@ -0,0 +1,325 @@
...
+     }
+     else
+     {
+          return (m_nan_value)(x - y);
+     }
+}
+
+m_nan_value m_nan_neg(m_nan_value x)
+{
+     return -x;
```
denismerigoux, line 52: Is is the case that `- NaN = Nan` ?

examples/c/m_nan_value.c
```diff
@@ -0,0 +1,325 @@
...
+}
+
+m_nan_value m_nan_mul(m_nan_value x, m_nan_value y)
+{
+     return x * y;
```
denismerigoux, line 57: Is it the case that `NaN` is an absorbant element of the float multiplication ?

examples/c/m_nan_value.c
```diff
@@ -0,0 +1,325 @@
...
+     else
+     {
+          double tmp = x + (x < 0 ? -0.50005 : 0.50005);
+          return (m_nan_value)(__int64_t)tmp;
+     }
+}
+
+m_nan_value m_nan_floor(m_nan_value x)
+{
+     return (m_nan_value)floor(x + 0.000001);
```
denismerigoux, line 249: What happens if `x` is `NaN` ?

src/mlang/backend_compilers/bir_to_c.ml
```diff
@@ -53,10 +55,10 @@ let generate_variable (var_indexes : int Mir.VariableMap.t) (offset : offset)
   | _ ->
       Format.fprintf fmt "TGV[%d/*%s*/%s]" var_index
         (Pos.unmark var.Mir.Variable.name)
-        ( match offset with
+        (match offset with
         | None -> ""
         | GetValue offset -> " + " ^ string_of_int offset
-        | PassPointer -> assert false )
```
denismerigoux, line 59: What is your `ocamlformat` version ? You can get is using `opam list ocamlformat`.

denismerigoux (approved):
Thanks for taking my suggestion into account! This `#define` trick will also be useful if we want to further experiment and switch C representations for M values.

This PR has quite a lot of noise due to an `ocamlformat`mismatch I think. I'm pushing a commit to this branch that reestablishes formatting uniformity.

Before merging, I would be sure that the `NaN` semantics do match the M `undef` semantics ; I've put a number of comments in `m_nan_value.c` and would be interested in confirming my assumptions.

dmentre:
Hello Denis,

Le 15/02/2021 à 17:14, Denis Merigoux a écrit :
>
> Before merging, I would be sure that the |NaN| semantics do match the 
> M |undef| semantics ; I've put a number of comments in |m_nan_value.c| 
> and would be interested in confirming my assumptions.
>
When doing the implementation, I used following program to make some tests:

-----------8<---------------
#include <math.h>
#include <stdio.h>

int main(void)
{
   double x;
   double tmp;

   printf("1.0 < NAN: %d\n", 1.0 < NAN);
   printf("1.0 >= NAN: %d\n", 1.0 >= NAN);
   printf("1.0 == NAN: %d\n", 1.0 == NAN);
   printf("1.0 && NAN: %d\n", 1.0 && NAN);
   printf("1.0 || NAN: %d\n", 1.0 || NAN);
   printf("NAN || 1.0: %d\n", NAN || 1.0);
   printf("1.0 + NAN: %f\n", 1.0 + NAN);
   printf("1.0 - NAN: %f\n", 1.0 - NAN);
   printf("1.0 / NAN: %f\n", 1.0 / NAN);
   printf("1.0 * NAN: %f\n", 1.0 * NAN);
   printf("!NAN: %f\n", (double)!NAN);
   printf("-NAN: %f isnan(-NAN): %d\n", -NAN, isnan(-NAN));
   printf("fmax(-1.0, NAN): %f\n", fmax(-1.0, NAN));
   printf("fmax(NAN, 1.0): %f\n", fmax(NAN, 1.0));
   printf("fmin(-1.0, NAN): %f\n", fmin(-1.0, NAN));
   x=1.3; tmp= (x + (x < 0 ? -0.50005 : 0.50005));
   printf("x=1.3, x + (x < 0 ? -0.50005 : 0.50005) : %f\n",
          (double)(__int64_t)tmp);
   x=NAN; tmp= (x + (x < 0 ? -0.50005 : 0.50005));
   printf("x=NAN, x + (x < 0 ? -0.50005 : 0.50005) : %f\n",
          (double)(__int64_t)tmp);
   printf("floor(NAN + 0.000001): %f\n", floor(NAN + 0.000001));
   printf("floor(1.3 + 0.000001): %f\n", floor(1.3 + 0.000001));
}
-----------8<---------------

It produces following output:

1.0 < NAN: 0
1.0 >= NAN: 0
1.0 == NAN: 0
1.0 && NAN: 1
1.0 || NAN: 1
NAN || 1.0: 1
1.0 + NAN: nan
1.0 - NAN: nan
1.0 / NAN: nan
1.0 * NAN: nan
!NAN: 0.000000
-NAN: -nan isnan(-NAN): 1
fmax(-1.0, NAN): -1.000000
fmax(NAN, 1.0): 1.000000
fmin(-1.0, NAN): -1.000000
x=1.3, x + (x < 0 ? -0.50005 : 0.50005) : 1.000000
x=NAN, x + (x < 0 ? -0.50005 : 0.50005) : -9223372036854775808.000000
floor(NAN + 0.000001): nan
floor(1.3 + 0.000001): 1.000000

While it gives some hints, it might not be the guarantee you want. I can 
check with respect to IEEE 754 standard, but it will take me more time.

I also wanted to make some unit tests to check m_nan_value.c primitives 
against m_value.c ones but I haven't done it yet.


> ------------------------------------------------------------------------
>
> In examples/c/m_nan_value.c 
> <https://github.com/MLanguage/mlang/pull/15#discussion_r576293679>:
>
> > +     {
> +          return (m_nan_value)(-y);
> +     }
> +     else if (M_NAN_UNDEFINED(y))
> +     {
> +          return (m_nan_value)x;
> +     }
> +     else
> +     {
> +          return (m_nan_value)(x - y);
> +     }
> +}
> +
> +m_nan_value m_nan_neg(m_nan_value x)
> +{
> +     return -x;
>
> Is is the case that |- NaN = Nan| ?
>
-NaN leads -NaN. But this -NaN is recognized as NaN by isnan().

-NAN: -nan isnan(-NAN): 1

> ------------------------------------------------------------------------
>
> In examples/c/m_nan_value.c 
> <https://github.com/MLanguage/mlang/pull/15#discussion_r576294046>:
>
> > +          return (m_nan_value)x;
> +     }
> +     else
> +     {
> +          return (m_nan_value)(x - y);
> +     }
> +}
> +
> +m_nan_value m_nan_neg(m_nan_value x)
> +{
> +     return -x;
> +}
> +
> +m_nan_value m_nan_mul(m_nan_value x, m_nan_value y)
> +{
> +     return x * y;
>
> Is it the case that |NaN| is an absorbant element of the float 
> multiplication ?
>
Yes.


> ------------------------------------------------------------------------
>
> In examples/c/m_nan_value.c 
> <https://github.com/MLanguage/mlang/pull/15#discussion_r576294732>:
>
> > +m_nan_value m_nan_round(m_nan_value x)
> +{
> +     if (M_NAN_UNDEFINED(x))
> +     {
> +          return m_nan_undefined;
> +     }
> +     else
> +     {
> +          double tmp = x + (x < 0 ? -0.50005 : 0.50005);
> +          return (m_nan_value)(__int64_t)tmp;
> +     }
> +}
> +
> +m_nan_value m_nan_floor(m_nan_value x)
> +{
> +     return (m_nan_value)floor(x + 0.000001);
>
> What happens if |x| is |NaN| ?
>
floor(NAN + 0.000001): nan

> ------------------------------------------------------------------------
>
> In src/mlang/backend_compilers/bir_to_c.ml 
> <https://github.com/MLanguage/mlang/pull/15#discussion_r576296471>:
>
> >          | None -> ""
>           | GetValue offset -> " + " ^ string_of_int offset
> -        | PassPointer -> assert false )
>
> What is your |ocamlformat| version ? You can get is using |opam list 
> ocamlformat|.
>
0.16.0

Best regards,
david

denismerigoux:
> I can check with respect to IEEE 754 standard, but it will take me more time.

> I also wanted to make some unit tests to check m_nan_value.c primitives
against m_value.c ones but I haven't done it yet.

I think we can wait for both of these items. Indeed, while the test program you wrote for `NaN` behavior seems sound, what happens down at assembly or hardware level can depend on which compiler or ISA is used. So I would prefer that the behavior is checked with respect to the IEEE754 standard.

Related : I know the DGFiP runs their code on a mainframe with [GCOS7](https://en.wikipedia.org/wiki/General_Comprehensive_Operating_System). I think they said at some point that their float encoding is not IEEE754... I'll ask them again.

denismerigoux:
Ok, I got an answer from the DGFiP. They do use a Bull DPS 7000 running GCOS7 in production, and the encoding of floating point numbers is not IEEE754. However, we have a spec of this special floating point number format : https://support.bull.com/ols/product/system/gcos7/gcos7-com/g7-dps7000/doc-com/docf/g/47X257TN30-aug2011/47A260UL06.pdf#page=158.

Another info : the DGFiP is migrating to a newer mainframe in a few years, and this newer mainframe will run Z/OS and uses IEEE754. So it might not be useful to put in a lot of work to accommodate the Bull DSP7000 floats since they will soon become obsolete.

dmentre:
Hello Denis,

Le 16/02/2021 à 11:16, Denis Merigoux a écrit :
>
>     I can check with respect to IEEE 754 standard, but it will take me
>     more time.
>
>     I also wanted to make some unit tests to check m_nan_value.c
>     primitives
>     against m_value.c ones but I haven't done it yet.
>
> I think we can wait for both of these items. Indeed, while the test 
> program you wrote for |NaN| behavior seems sound, what happens down at 
> assembly or hardware level can depend on which compiler or ISA is 
> used. So I would prefer that the behavior is checked with respect to 
> the IEEE754 standard.
>
OK, then I'll check against IEEE 754 or a source close enough to IEEE 
754 (as the standard is sold for 104$ and this is a hobby project for 
me). I'll also make unit tests against ml_value.c primitives.


> Ok, I got an answer from the DGFiP. They do use a Bull DPS 7000 
> running GCOS7 in production, and the encoding of floating point 
> numbers is not IEEE754. However, we have a spec of this special 
> floating point number format : 
> https://support.bull.com/ols/product/system/gcos7/gcos7-com/g7-dps7000/doc-com/docf/g/47X257TN30-aug2011/47A260UL06.pdf#page=158 
> <https://support.bull.com/ols/product/system/gcos7/gcos7-com/g7-dps7000/doc-com/docf/g/47X257TN30-aug2011/47A260UL06.pdf#page=158>.
>
> Another info : the DGFiP is migrating to a newer mainframe in a few 
> years, and this newer mainframe will run Z/OS and uses IEEE754. So it 
> might not be useful to put in a lot of work to accommodate the Bull 
> DSP7000 floats since they will soon become obsolete.
>
I agree with you that it is not worth making efforts regarding Bull DPS 
7000 system and focus on IEEE 754.

But even for IEEE 754 compliant hardware, I've been told that effective 
support might be different from hardware to hardware. One approach for 
this would be to develop a "qualification suite", i.e. tests that would 
check that a given hardware/compiler combination has the expected 
semantics. This is the approach used but ANSYS/Esterel Technologies for 
KCG, their SCADE compiler.

Best regards,
david

dmentre:
Hello Denis,

Quick update on current status:
- **IEEE 754 standard**: I looked briefly at it but it is too abstract to be understandable. I have not time and willing to dig into it
- **unit tests**: I started to write a small framework and done some tests. But I'm not satisfied at all. Choosing the right test cases is time consuming. Even with current quite arbitrary values, some tests are not working and I haven't investigated to understand why. I also don't have the proper tooling (testing tool, test generation tool, coverage measurement tool) to do a proper work. And debugging C code is not my favourite leisure activity. :-( I nonetheless pushed my changes if somebody is willing to start from it or use inspiration from it.

Best regards,
david

Issue #16, opened 2021-07-20
Static binary compilation for older platforms


jemsab:
In order to run the main binary on older linux platforms, the binary must be statically linked with the musl implementation of the C standard library. 

Unfortunately the gmp dependency seems to prevent such a linking to take place using an opam compiler with musl and static enabled.

Indeed `make deps` gives the following error output : 
opam install ppx_deriving ANSITerminal re ocamlgraph dune menhir \
	cmdliner dune-build-info visitors parmap num ocamlformat mlgmpidl \
	ocamlformat
[NOTE] Package ocamlformat is already installed (current version is 0.18.0).
[NOTE] Package ocamlformat is already installed (current version is 0.18.0).
[NOTE] Package num is already installed (current version is 1.4).
[NOTE] Package parmap is already installed (current version is 1.2.3).
[NOTE] Package visitors is already installed (current version is 20210608).
[NOTE] Package dune-build-info is already installed (current version is 2.9.0).
[NOTE] Package cmdliner is already installed (current version is 1.0.4).
[NOTE] Package menhir is already installed (current version is 20210419).
[NOTE] Package dune is already installed (current version is 2.9.0).
[NOTE] Package ocamlgraph is already installed (current version is 2.0.0).
[NOTE] Package re is already installed (current version is 1.9.0).
[NOTE] Package ANSITerminal is already installed (current version is 0.8.2).
[NOTE] Package ppx_deriving is already installed (current version is 5.2.1).
The following actions will be performed:
  ∗ install mlgmpidl 1.2.13
Do you want to continue? [Y/n] y

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[mlgmpidl.1.2.13] found in cache

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ERROR] The compilation of mlgmpidl failed at
        "/home/jbarnes01/.opam/opam-init/hooks/sandbox.sh build make".

#=== ERROR while compiling mlgmpidl.1.2.13 ====================================#
\# context     2.0.7 | linux/x86_64 | ocaml-option-flambda.1 ocaml-option-musl.1 ocaml-option-static.1 ocaml-variants.4.12.0+options | https://opam.ocaml.org/#fde0c1b5
\# path        ~/.opam/4.12.0+flambda+static/.opam-switch/build/mlgmpidl.1.2.13
\# command     ~/.opam/opam-init/hooks/sandbox.sh build make
[...]

PR #27, opened 2021-07-26
static make

jemsab:
Solution for issue #16

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,10 @@ format:
 build: format
 	dune build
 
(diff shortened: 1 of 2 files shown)


jemsab:
This basic solution could be expanded to include a Dockerfile and Github action in order to deploy static builds that can be downloaded straight for Github.

denismerigoux (approved):
Thanks James, merging that right away!

denismerigoux:
Closes #16

PR #46, opened 2021-07-28
Header files

### LINKED ISSUE #45: C backend header files problems
The m_value.h header file does not include a header guard which could lead to a double definition compiler error for the m_value struct when including both ir_xxxx.h and m_value.h.

The ir_xxx.h file includes m_value.h inside <> rather than "" which can also cause a compiler error when compiling these files without using the Mlang makefile.

jemsab:
closes #45

### DIFF
--- a/examples/c/m_value.h
+++ b/examples/c/m_value.h
@@ -1,3 +1,6 @@
+#ifndef M_VALUE_
+#define M_VALUE_
+
 #include <math.h>
 #include <stdbool.h>
 #include <stdlib.h>
@@ -39,3 +42,5 @@ bool m_is_defined_false(m_value x);
 m_value m_literal(double v);
 m_value m_array_index(m_value *array, m_value index, int array_size);
 m_value m_multimax(m_value bound, m_value *array);
+
+#endif /* M_VALUE_ */ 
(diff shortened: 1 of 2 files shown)


rmonat:
It seems the CI is not triggered correctly. I can investigate tomorrow if needed.

jemsab:
Is this a main branch vs other branch issue ? 

The main branch seems to be pulled by gitlab without any difficulties but here we get a 403.
I think we may indeed need your help as I don't have access to the gitlab clone.

rmonat:
This might be because the branch is not in `MLanguage/mlang` but on your repository. Sorry about that. Could you check by creating a branch on this repo?

jemsab:
No problems, thanks for your help! I've just pushed a gitlab-test-branch branch

jemsab:
lgtm https://gitlab.com/rmonat/mlang/-/pipelines/344702229

rmonat:
Good! Then I suggest prototyping on branches of this repository directly (if that doesn't break a workflow you agreed upon with the rest of the team?).

denismerigoux:
Thanks for this PR James! This indeed solves #45  problem that I had encountered before. I don't have much C development experience and didn't know that was how you solved this annoying double inclusion problem (related to C's horrible module system).

Issue #19, opened 2021-07-22
Create new DGFiP backend


jemsab:
Create new Mlang backend:

- Keep current C backend and add another one that's specific to the DGFiP. 
- Write an M++ file containing a single function that only calls call_m once.

PR #50, opened 2021-08-09
remove anomaly level limit

### LINKED ISSUE #24: Implement calculation anomaly handling
Implement calculation anomaly handling as defined in the M code, with every level of anomaly.

Give a way to configure what happens for each anomaly level with C compilation flags (implement the same solution as the current DGFiP implementation).

jemsab:
closes #24 

Should this PR include a change of the current C backend or should we have a preliminary version of the DGFiP specific C backend ?

### DIFF
--- a/src/mlang/backend_compilers/bir_to_c.ml
+++ b/src/mlang/backend_compilers/bir_to_c.ml
@@ -162,28 +162,33 @@ let generate_var_def (var_indexes : int Mir.VariableMap.t) (var : Mir.Variable.t
 
 let generate_var_cond (var_indexes : int Mir.VariableMap.t) (cond : condition_data)
     (oc : Format.formatter) =
-  let scond, defs = generate_c_expr cond.cond_expr var_indexes in
-  let percent = Re.Pcre.regexp "%" in
-  Format.fprintf oc
-    "%acond = %s;@\n\
-     if (m_is_defined_true(cond)) {@\n\
-    \    printf(\"Error triggered: %a\\n\");@\n\
-    \    {@\n\
-    \        output->is_error = true;@\n\
-    \        free(TGV);@\n\
-    \        free(LOCAL);@\n\
-    \        return -1;@\n\
-    \    }@\n\
-     }@\n"
-    (format_local_vars_defs var_indexes)
-    defs scond
-    (Format.pp_print_list
-       ~pp_sep:(fun fmt () -> Format.fprintf fmt "; ")
-       (fun fmt err ->
-         let error_descr = Pos.unmark err.Error.descr in
-         let error_descr = Re.Pcre.substitute ~rex:percent ~subst:(fun _ -> "%%") error_descr in
-         Format.fprintf fmt "%s: %s" (Pos.unmark err.Error.name) error_descr))
-    cond.cond_errors
+  if
+    List.fold_left
+      (fun acc (item : Error.t) -> if acc = true then true else item.typ = Mast.Anomaly)
+      false cond.cond_errors
+  then
+    let scond, defs = generate_c_expr cond.cond_expr var_indexes in
+    let percent = Re.Pcre.regexp "%" in
+    Format.fprintf oc
+      "%acond = %s;@\n\
+       if (m_is_defined_true(cond)) {@\n\
+      \    printf(\"Error triggered: %a\\n\");@\n\
+      \    {@\n\
+      \        output->is_error = true;@\n\
+      \        free(TGV);@\n\
+      \        free(LOCAL);@\n\
+      \        return -1;@\n\
+      \    }@\n\
+       }@\n"
+      (format_local_vars_defs var_indexes)
+      defs scond
+      (Format.pp_print_list
+         ~pp_sep:(fun fmt () -> Format.fprintf fmt "; ")
+         (fun fmt err ->
+           let error_descr = Pos.unmark err.Error.descr in
+           let error_descr = Re.Pcre.substitute ~rex:percent ~subst:(fun _ -> "%%") error_descr in
+           Format.fprintf fmt "%s: %s" (Pos.unmark err.Error.name) error_descr))
+      cond.cond_errors
 
 let fresh_cond_counter = ref 0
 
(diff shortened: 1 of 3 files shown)


jemsab:
Started work on DGFiP specific C backend referenced by #19 in order to communicate a parametrized error list to calling layer. 

If we want to "upstream" this to the historic C backend it shouldn't be too hard.

denismerigoux:
The problem with this PR is that the current test suite raises a number of "anomalies" with low-levels (informative), that will generate exit errors in the current C backend. So actually this commit is already breaking the C backend; we don't see it here because you made this PR from your fork instead of doing it from this repo (@rmonat we should find a solution for these CI issues ?). 

What you should do to fix that is add an anomaly level to the MIR data structure that stores verification conditions:

https://github.com/MLanguage/mlang/blob/3e2cf2b376f2cced6fd049b5d84d0d7b2ee2a1a1/src/mlang/m_ir/mir.ml#L271

Then you can propagate this information down to BIR and choose in the current C backend (`bir_to_c.ml`) to not emit verification conditions with an anomaly level lower than "error". That way it'll be cleaner and you'll have what you need for the DGFiP backend.

jemsab:
> The problem with this PR is that the current test suite raises a number of "anomalies" with low-levels (informative), that will generate exit errors in the current C backend. So actually this commit is already breaking the C backend; we don't see it here because you made this PR from your fork instead of doing it from this repo (@rmonat we should find a solution for these CI issues ?).
> 

Raphaël previously suggested we move all development work to this project rather having our own forks. What do you think Denis? Another solution would be turn on the tests in Github actions, as I understand you had it working until b7b3b0ea9d98701fa3fba7742f5b35f22db08a66

> What you should do to fix that is add an anomaly level to the MIR data structure that stores verification conditions:
> 
> https://github.com/MLanguage/mlang/blob/3e2cf2b376f2cced6fd049b5d84d0d7b2ee2a1a1/src/mlang/m_ir/mir.ml#L271
> 
> Then you can propagate this information down to BIR and choose in the current C backend (`bir_to_c.ml`) to not emit verification conditions with an anomaly level lower than "error". That way it'll be cleaner and you'll have what you need for the DGFiP backend.

Ok I'll start with those modifications, thanks!

denismerigoux:
> Raphaël previously suggested we move all development work to this project rather having our own forks. What do you think Denis? 

Please do, all the people working on this now have write acces on this repo.

> Another solution would be turn on the tests in Github actions, as I understand you had it working until b7b3b0e

Unfortunately the CI we have is really comprehensive, testing all cases on all backends and this would overshoot GitHub Action's free plan limit. That is why we use the machine in Raphaël's lab. But we can configurate better the link with Raphaël's machine I think, we'll have to look into that.

denismerigoux:
By the way you can replicate what the CI would do on your local machine with the `make all` command (which includes `make tests`, `make test_c_backend`, ... Anyway it seems that the CI started, we'll see the results soon :)

jemsab:
As you predicted, a total failure, at least CI works this way which is a definite improvement!

jemsab:
> > Another solution would be turn on the tests in Github actions, as I understand you had it working until [b7b3b0e](https://github.com/MLanguage/mlang/commit/b7b3b0ea9d98701fa3fba7742f5b35f22db08a66)
> 
> Unfortunately the CI we have is really comprehensive, testing all cases on all backends and this would overshoot GitHub Action's free plan limit. That is why we use the machine in Raphaël's lab. But we can configurate better the link with Raphaël's machine I think, we'll have to look into that.

Perhaps some docker magic with multistage builds would allow us to improve the test run times, either on gitlab or here.

jemsab:
As I don't have enough RAM on my machine to compile the C tests file, I have to use the CI server for the moment. For the C engine at least. 

As to changing the Mir tree, `cond_errors` already carries the anomaly type through inside `Error.t`, would it not be acceptable to use that instead? The latest commit is a test with that supposition in any case.

denismerigoux:
Oh nicely spotted, I had forgotten I had already stored the anomaly level in `Error.t` :)

src/mlang/backend_compilers/bir_to_c.ml
```diff
@@ -162,28 +162,33 @@ let generate_var_def (var_indexes : int Mir.VariableMap.t) (var : Mir.Variable.t
...
-    (Format.pp_print_list
-       ~pp_sep:(fun fmt () -> Format.fprintf fmt "; ")
-       (fun fmt err ->
-         let error_descr = Pos.unmark err.Error.descr in
-         let error_descr = Re.Pcre.substitute ~rex:percent ~subst:(fun _ -> "%%") error_descr in
-         Format.fprintf fmt "%s: %s" (Pos.unmark err.Error.name) error_descr))
-    cond.cond_errors
+  if
+    List.fold_left
+      (fun acc (item : Error.t) -> if acc = true then true else item.typ = Mast.Anomaly)
```
denismerigoux, line 167: This is `List.exists` (https://ocaml.org/api/List.html)

jemsab: Ah ha, you learn something every day!

jemsab: Once this current round of CI confirms that the problem is solved, I'll change both uses of List.fold_left

Keryan-dev:
> The problem with this PR is that the current test suite raises a number of "anomalies" with low-levels (informative), that will generate exit errors in the current C backend.

As I understood, only Anomaly level errors must generate exit errors. Is there a deeper reason to not want other level handled appropriately (as in, do what it was meant to do) in the current backend, regardless of the need to update the testsuite (maybe that is also a lot of work?) ?

denismerigoux:
Hi Keryan, to answer your question, there is deeper reason, and yes updating the fuzzer-generated test suite to take non-error anomalies into account would be a lot of work (probably something like one week since you have to run the fuzzer again).

jemsab:
I would also add that error handling is decided by the client. Some prevent further calculation after 10 Anomaly level errors, some go through the whole calculation. 

I do wonder if we should hang on to that mechanism. As I understand it, it is supposed to be a performance optimisation based on the 0.5 microseconds that might be saved, it's probably not worth it. With Amdahl's Law, there are probably far better optimizations elsewhere in any case.

PR #53, opened 2021-08-11
Anomaly handling dgfip backend

jemsab:
dummy pr until dgfip backend creation can be merged so as to have ci


denismerigoux:
When you create a PR, you need to to merge it into master. You can specify another branch to merge that into. Here, you should specify that you want to merge that anomaly branch into the specific DGFiP backend branch.

jemsab:
> When you create a PR, you need to to merge it into master. You can specify another branch to merge that into. Here, you should specify that you want to merge that anomaly branch into the specific DGFiP backend branch.

Oh sorry, I've perhaps made a second mistake as well, as I wasn't sure if and when `dgfip-backend` would get merged into master as is, this branch is an offshoot of anomaly handling with `dgfip-backend` mixed in.

My idea was that once `dgfip-backend` is merged into master at some point, I can merge this branch back into `anomaly-handling`, and then merge `anomaly-handling` into master with the original PR. The idea being to keep `anomaly-handling` mergeable at any point whatever the state of `dgfip-backend` is. 

I only created this PR to get the results of CI testing on my changes as despite my 16GB of RAM on this machine, compiling the C backends occasionally crashes the whole OS as it starts swapping like mad.

denismerigoux:
Yes it's a good idea to use PRs as a way to get CI. But Github PRs are very flexible, and as I was saying when you create one you can choose `dgfip-backend` as the merge target instead of `master`; that way your merging strategy is simpler. The advantage of doing this is code review: until `dgfip-backend` is not merged in `master`, the diff of this PR is huge (1700 additions) and it's difficult for me to know which things this PR adds specifically VS the things `dgfip-backend` adds. If you create PRs with small diffs by merging directly to the closest branch, I can review more easily and give you more relevant comments.

jemsab:
Thank you for explaining, it's much clearer for now, I've changed the base branch to `dgfip-backend`

jemsab:
An infinite loop seems to be occurring inside test_harness.exe for the DGFiP specific backend, with the loop 

```c
while ((dir = readdir(d)) != NULL) {
// handle test case
}
```

which is strictly the same as the test_harness used by the C backend. 

The variable `dir` is a non null struct each time with an empty but defined string for d_name. 

The problem is reproducible on my machine as well as the CI server. I'll keep looking but for the moment this has me stumped...

jemsab:
Commenting out the body of the loop eliminated the infinite loop, so the problem is most likely somewhere in ir_tests.c of the new backend.

src/mlang/backend_compilers/bir_to_dgfip_c.ml
```diff
@@ -15,6 +15,20 @@ open Mir
 
 let none_value = "m_undefined"
 
+module Error = struct
+  type t = {
```
denismerigoux, line 19: This type should actually replace `Mir.Error`, because I didn't realize that the list of strings has actually a strict meaning.

examples/dgfip_c/backend_tests/test_harness.c
```diff
@@ -86,7 +89,10 @@ int main(int argc, char *argv[])
                         // Here we move to controlling the outputs, so we
                         // have to run the computation!
                         m_input_from_array(input_for_m, input_array_for_m);
+                        errors = malloc(sizeof(Errors));
```
denismerigoux, line 92: I still can't see where this `Errors` live, it does not appear to be generated in `bir_to_dgfip_c.ml`.

denismerigoux (changes requested):
Very nice! I left a few comments for details. However before we merge this we should add to the testing harness a way to check expected errors in test files.

examples/dgfip_c/backend_tests/test_harness.c
jemsab: As mentioned IRL, Errors is typedef of an m_error table of the size of the number of errors possible during a calculation.

src/mlang/m_ir/mir.ml
```diff
@@ -246,11 +246,22 @@ type variable_data = {
 (**{1 Verification conditions}*)
 
 (** Errors are first-class objects *)
+
 module Error = struct
+  module ErrorDesc = struct
+    type t = {
```
denismerigoux, line 252: You don't need to put it inside another module, just write `type error_descr = ...`

denismerigoux (approved):
Thanks for the changes! You can implement the small change I propose and go ahead and merge :)

Issue #26, opened 2021-07-23
Slice back-end code in functions corresponding to M rules


Keryan-dev:
This is the first item of issue #22, which deserve its own issue I think.

The goal is to generate functions that reflect the rules of input M specs.
Since this will have a negative impact on performance due to the restrictions on optimisation passes, it should have its own compilation flags.

Keryan-dev:
I can't assign myself to an issue (yet ?), so I do so verbally here.

denismerigoux:
Hi Keryan, thanks for creating the issue. I'm not sure will have a negative impact on performance, what restrictions on the optimisation passes do you think it will create ? For me, this will just transform `type program = stmt list` into `type program = rule list` where `type rule = stmt list`, no ?

Keryan-dev:
> I'm not sure will have a negative impact on performance, what restrictions on the optimisation passes do you think it will create ? For me, this will just transform `type program = stmt list` into `type program = rule list` where `type rule = stmt list`, no ?

That's the plan, yes. You know the optimisation passes better than I do, but I was under the impression that it takes advantage of the fact that all the statements of a program are grouped together to optimise a lot through reordering (among other things), which would not be possible if statements are spread around in small blocks of codes that can only be optimised separately.

denismerigoux:
Closed by #55

PR #54, opened 2021-08-18
Split M code in rules in pipeline without optimisations

Keryan-dev:
This addresses issue #26.

For now it only applies to unoptimized compilation pipeline, and only changes the C backend.

For a reason yet unknown, it decreases efficiency and performance of the optimisations passes (yes, even without actually changing things). I hope a review could give some insight on this.


Keryan-dev:
Closing since PR is now open from this repo.

PR #55, opened 2021-08-19
Split M code in rules

Keryan-dev:
This addresses issue #26.

For now it only applies to unoptimized compilation pipeline, and only changes the C backend.

For a reason yet unknown, it decreases efficiency and performance of the optimisations passes (yes, even without actually changing things). I hope a review could give some insight on this.

### DIFF
--- a/src/mlang/backend_compilers/bir_to_c.ml
+++ b/src/mlang/backend_compilers/bir_to_c.ml
@@ -36,7 +36,7 @@ let generate_binop (op : Mast.binop) : string =
 let generate_unop (op : Mast.unop) : string =
   match op with Mast.Not -> "m_not" | Mast.Minus -> "m_neg"
 
-type offset = GetValue of int | PassPointer | None
+type offset = GetValueConst of int | GetValueVar of string | PassPointer | None
 
 let generate_variable (var_indexes : int Mir.VariableMap.t) (offset : offset)
     (fmt : Format.formatter) (var : Variable.t) : unit =
@@ -55,7 +55,8 @@ let generate_variable (var_indexes : int Mir.VariableMap.t) (offset : offset)
         (Pos.unmark var.Mir.Variable.name)
         (match offset with
         | None -> ""
-        | GetValue offset -> " + " ^ string_of_int offset
+        | GetValueVar offset -> " + " ^ offset
+        | GetValueConst offset -> " + " ^ string_of_int offset
         | PassPointer -> assert false)
 
 let generate_raw_name (v : Variable.t) : string =
@@ -117,7 +118,7 @@ let rec generate_c_expr (e : expression Pos.marked) (var_indexes : int Mir.Varia
   | Literal Undefined -> (Format.asprintf "%s" none_value, [])
   | Var var -> (Format.asprintf "%a" (generate_variable var_indexes None) var, [])
   | LocalVar lvar -> (Format.asprintf "LOCAL[%d]" lvar.LocalVariable.id, [])
-  | GenericTableIndex -> (Format.asprintf "generic_index", [])
+  | GenericTableIndex -> (Format.asprintf "m_literal(generic_index)", [])
   | Error -> assert false (* should not happen *)
   | LocalLet (lvar, e1, e2) ->
       let _, s1 = generate_c_expr e1 var_indexes in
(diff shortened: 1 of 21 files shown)


src/mlang/backend_compilers/bir_to_c.ml
```diff
@@ -153,11 +153,16 @@ let generate_var_def (var_indexes : int Mir.VariableMap.t) (var : Mir.Variable.t
...
-      Errors.raise_spanned_error "generic index table definitions not supported in C the backend"
-        (Pos.get_position e)
+  | TableVar (_size, IndexGeneric _e) ->
+      (* Format.asprintf
+       *   "for (int generic_index=0; generic_index < %d; generic_index++) {@\n\
+       *   \ @[<h\n\
+       *   \         4> %a = %a;@]@\n\
+       *   \ }@\n"
+       *   size generate_variable var generate_c_expr e *)
+      Format.fprintf oc "/* table def*/"
```
denismerigoux, line 163: Ok, this is the WIP you talked to me about

src/mlang/backend_ir/bir_interpreter.ml
```diff
@@ -524,6 +524,9 @@ module Make (N : Bir_number.NumberInterface) = struct
         match evaluate_expr ctx p.mir_program data.cond_expr with
         | Number f when not (N.is_zero f) -> report_violatedcondition data ctx
         | _ -> ctx)
+    | Bir.SRuleCall _ ->
+        (* removed with [Bir.get_all_statements] below *)
```
denismerigoux, line 528: It would be better for debug purposes to actually interpret rules one by one rather than bundling everything in one list of statements.

src/mlang/backend_ir/bir.ml
```diff
@@ -12,115 +12,118 @@
    You should have received a copy of the GNU General Public License along with this program. If
    not, see <https://www.gnu.org/licenses/>. *)
 
-type stmt = stmt_kind Pos.marked
+type rule_id = int
```
denismerigoux, line 15: `type rule_id = Mir.rule_id` ?

src/mlang/m_ir/mir.ml
```diff
@@ -287,6 +306,7 @@ type exec_pass = { exec_pass_set_variables : literal Pos.marked VariableMap.t }
 
 type program = {
   program_vars : variable_data VariableMap.t;
+  program_rules : rule_data RuleMap.t;
```
denismerigoux, line 309: I would prefer this type to reflect more the structure of the programs with only one rule map where each value contains a little map of the variables defined inside this rule.

src/mlang/m_ir/mir_dependency_graph.ml
```diff
@@ -57,20 +67,36 @@ let add_usages (lvar : Mir.Variable.t) (e : Mir.expression Pos.marked) (acc : G.
   let usages = get_used_variables_ e Mir.VariableMap.empty in
```
denismerigoux, line 67: So here keeping only one dependency graph, in which rules are nodes, would be better. The edges in ocamlgraph can also carry labels. Here the labels would be the variables inside the rules that create dependencies. See this example https://github.com/CatalaLang/catala/blob/master/compiler/scopelang/dependency.ml for how to create labels on edges and how using them to display good error messages for cycle detection.

src/mlang/m_ir/mir.ml
denismerigoux: In fact, inside a given rule, we should store a list of variable assignations and not a map, since the order is important.

src/mlang/optimizing_ir/bir_to_oir.ml
```diff
@@ -28,18 +28,14 @@ let initialize_block (bid : Oir.block_id) (blocks : Oir.block Oir.BlockMap.t) :
...
-    (blocks : Oir.block Oir.BlockMap.t) : Oir.block_id * Oir.block Oir.BlockMap.t =
-  let blocks, last_block_id =
-    List.fold_left
-      (fun (blocks, current_block_id) stmt ->
-        let new_current_block_id, new_blocks = translate_statement stmt current_block_id blocks in
-        (new_blocks, new_current_block_id))
-      (blocks, curr_block_id) l
-  in
-  (last_block_id, blocks)
+let rec translate_statement_list (p : Bir.program) (l : Bir.stmt list)
```
denismerigoux, line 31: `Oir` should contain the concept of rule. Basic blocks in `Oir` should consist of a list of rules that are chained in the control flow.

denismerigoux (changes requested):
Thanks very much! We've discussed the changes IRL, but I left comments as a reminder :)

Keryan-dev:
I just committed something for optimisation passes. Although I don't expect the CI to succeed, I obtained *better* results on the first pass of dead code elimination which leads me to think there is something very wrong, but I can't figure out what from my relatively local changes.

Keryan-dev:
Oh, the Python backend seems to raise helpful errors at least, I have my clue then.

src/mlang/mpp_ir/mpp_ir_to_bir.ml
```diff
@@ -10,11 +10,27 @@ end)
...
+  {
+    new_variables = StringMap.empty;
+    variables_used_as_inputs = Mir.VariableDict.empty;
+    rule_instances = Bir.RuleMap.empty;
+  }
+
+let ctx_join ctx1 ctx2 =
+  {
+    new_variables =
+      (* CR keryan : I assume there shouldn't be conflict there, kept the behavior of
```
denismerigoux, line 27: Indeed, there shouldn't be any conflict. What happens if you replace `Some v2` by `assert false` and run the tests again ?

denismerigoux (approved):
Looks like you've succeeded :) You can merge now or after the little test I proposed.

denismerigoux:
Closes #26

src/mlang/mpp_ir/mpp_ir_to_bir.ml
Keryan-dev: By conflict I meant cases where `v1` and `v2` would be different, but using this function is unavoidable.
`ctx_join` is used to merge contexts of the branches of conditional expression, each branch is evaluated with the same entry context, including `new_variables` whose intersection will most likely not be empty.

Keryan-dev: So the test would be `assert (Mir.Variable.compare v1 v2 = 0)` (assuming comparing the ids is enough), and there is indeed no issues on the tests.

Keryan-dev:
Side note: on my laptop, the `ulimit` command on the C backend test fails (the value is too high, I assume) but the test manage to run nonetheless.

PR #57, opened 2021-08-23
 input output configuration for m_spec 

### LINKED ISSUE #20: Modify input-out configuration system
Modify input-out configuration system used by Mlang, based on the .m_spec files, so as to easily select the input/outputs expected by the C source files of higher layers of DGFiP code (INTER files).

Add keywords to allow adding arbitrary number of variables, for exemple : allow all available inputs or all variables with `cotsoc` is 1

jemsab:
closes #20


src/mlang/m_ir/mir.ml
```diff
@@ -359,3 +359,8 @@ let find_var_by_name (p : program) (name : string Pos.marked) : Variable.t =
            (fun v1 v2 -> compare v1.Variable.execution_number v2.Variable.execution_number)
            (Pos.VarNameToID.find name p.program_idmap))
     with Not_found -> Errors.raise_spanned_error "unknown variable" (Pos.get_position name))
+
+let find_vars_by_io (p : program) (io_to_find : io) : Variable.t list =
+  VariableMap.fold
+    (fun k v acc -> if v.var_io = io_to_find then k :: acc else acc)
+    p.program_vars []
```
Keryan-dev, line 366: Unless you have a reason to keep variables in a certain order, I would suggest to keep it as a map.
You have `Map.filter` for this, or in your case `Map.filter_map` since you discard the value.

src/mlang/m_ir/mir.ml
```diff
@@ -411,7 +411,53 @@ let find_var_by_name (p : program) (name : string Pos.marked) : Variable.t =
...
-  VariableMap.fold
-    (fun k v acc -> if v.var_io = io_to_find then k :: acc else acc)
-    p.program_vars []
+let find_vars_by_io (p : program) (io_to_find : io) : rule_data RuleMap.t =
+  let is_var_io (_, var_data) = var_data.var_io = io_to_find in
+  RuleMap.filter_map
+    (fun _ y ->
+      let rule_vars = List.filter is_var_io y.rule_vars in
+      Some { y with rule_vars })
+    p.program_rules
```
Keryan-dev, line 420: I would expect this function to return a `VarDict.t`, as the name is similar to the other helper function of this kind.
If you actually need other information of `var_definition`, I think the name would need to change.

(But yes, dissociating `var_definition` from the actual variable map is cumbersome as we see in the next function below.)

src/mlang/m_ir/mir.ml
```diff
@@ -411,7 +411,53 @@ let find_var_by_name (p : program) (name : string Pos.marked) : Variable.t =
...
+
+let fold_vars (f : Variable.t -> variable_data -> 'a -> 'a) (p : program) (acc : 'a) : 'a =
+  RuleMap.fold
+    (fun _ rule_data acc ->
+      List.fold_left
+        (fun acc (vid, def) ->
+          let var = VariableDict.find vid p.program_vars in
+          f var def acc)
+        acc rule_data.rule_vars)
+    p.program_rules acc
```
Keryan-dev: Nitpicking: rebasing PR branch instead of merging target branch into it avoids bloating the diff.

Keryan-dev: Forget it, github showed me a weird version of the diff form some reason.

src/mlang/m_ir/mir.ml
```diff
@@ -411,6 +411,19 @@ let find_var_by_name (p : program) (name : string Pos.marked) : Variable.t =
            (Pos.VarNameToID.find name p.program_idmap))
     with Not_found -> Errors.raise_spanned_error "unknown variable" (Pos.get_position name))
 
+let find_vars_by_io (p : program) (io_to_find : io) : VariableDict.t =
+  let rules_as_list = RuleMap.bindings p.program_rules in
+  let rule_vars =
+    List.flatten (List.map (fun rd -> rd.rule_vars) (List.map (fun (_, rd) -> rd) rules_as_list))
```
Keryan-dev, line 417: Since we're now in 4.11, we have `List.concat_map` that combine `flatten` and `map` in an efficient (and tail-recursive !) way.
Also `(fun (_, rd) -> rd)` is effectively the pervasive function `snd`.

src/mlang/m_ir/mir.ml
```diff
@@ -411,6 +411,19 @@ let find_var_by_name (p : program) (name : string Pos.marked) : Variable.t =
...
+  in
+  let filtered_list =
+    List.filter (fun (_, y) -> y.var_io = io_to_find) rule_vars
+    |> List.map (fun (var_id, _) -> var_id)
```
Keryan-dev, line 421: Same here, `List.filter_map` is what you want.

src/mlang/m_ir/mir.ml
```diff
@@ -411,6 +411,16 @@ let find_var_by_name (p : program) (name : string Pos.marked) : Variable.t =
...
 
+let find_vars_by_io (p : program) (io_to_find : io) : VariableDict.t =
+  let rules_as_list = RuleMap.bindings p.program_rules in
+  let rule_vars =
+    List.concat_map (fun (_,rd) -> rd.rule_vars) rules_as_list in
+  let filtered_list =
+    List.filter_map  (fun (var_id, var_data) -> if var_data.var_io = io_to_find then Some var_id else None) rule_vars in
+  VariableDict.filter
+    (fun vid _ -> List.exists (fun list_vid -> list_vid = vid) filtered_list)
+    p.program_vars
```
jemsab, line 422: Refactor this function to use maps, so as to have line 421 Map.mem rather than List.exists

jemsab:
After disabling optimization passes, C backend can successfully produce files. Something is making the optimizer go overboard removing too many variables.

jemsab:
This PR passes tests but doesn't fully answer issue #20. 

Indeed, the issue specifies fine grain control over the input variables using in a calculation. The need for such a level of control does not appear to answer a specific need within the DGFiP ecosystem. This PR allows the creation of generic calculation engine that accepts all inputs and outputs.

I'm not sure whether we still want to keep issue #20 open, to be revisited at a later date, or whether we should close the issue now and reopen another should the need arise

src/mlang/backend_ir/bir_interface.ml
```diff
@@ -201,7 +233,17 @@ let adapt_program_to_function (p : Bir.program) (f : bir_function) : Bir.program
                     {
                       Mir.var_typ = None;
                       Mir.var_io = Regular;
-                      Mir.var_definition = Mir.SimpleVar (Mir.Literal Mir.Undefined, pos);
+                      Mir.var_definition =
```
denismerigoux, line 236: Good catch!

src/mlang/driver.ml
```diff
@@ -65,8 +65,12 @@ let driver (files : string list) (debug : bool) (var_info_debug : string list) (
...
     then Errors.raise_error "Cycles between rules.";
     let mpp = Mpp_frontend.process mpp_file full_m_program in
-    let m_program = Mir_interface.reset_all_outputs full_m_program.program in
-    let full_m_program = Mir_interface.to_full_program m_program in
+    let full_m_program =
+      Mir_interface.to_full_program
+        (match function_spec with
+        | Some _ -> Mir_interface.reset_all_outputs full_m_program.program
+        | None -> full_m_program.program)
+    in
```
denismerigoux, line 73: So if I understand correctly, the new behavior is that if you don't pas `--function_spec=...` to the CLI arguments of Mlang, then it behaves as the old DGFiP compiler, assuming everything can be output and input? If that's the case, I would like this to be documented in `src/cli.ml` in the `function_spec` option documentation.

src/mlang/m_frontend/mast_to_mir.ml
```diff
@@ -1216,10 +1216,16 @@ let add_dummy_definitions_for_variable_declarations (var_data : Mir.variable_dat
...
+                    begin
+                      match io with
+                      | Mir.Input -> Mir.InputVar
+                      | _ ->
+                          Mir.TableVar
+                            ( size,
+                              Mir.IndexGeneric
+                                (Pos.same_pos_as (Mir.Literal Mir.Undefined) var.Mir.Variable.name)
+                            )
+                    end;
```
denismerigoux, line 1228: This should be refactored into a `Mir.create_undefined_value : io -> value` function.

src/mlang/m_ir/mir.ml
```diff
@@ -538,3 +543,18 @@ let fold_vars (f : Variable.t -> variable_data -> 'a -> 'a) (p : program) (acc :
...
+
+let find_vars_by_io (p : program) (io_to_find : io) : VariableDict.t =
+  fold_vars
+    (fun var var_data acc ->
+      if
+        var_data.var_io = io_to_find
+        && var
+           = List.hd
+               (Pos.VarNameToID.find (Pos.unmark var.name) p.program_idmap
+               |> List.sort (fun v1 v2 ->
```
denismerigoux, line 555: OK that's kind of awkard but I see what you want to do. You want to remove the duplicates created by the phantom definitions of all variables to `undefined`. A better way to do this is simply to test whether the rule number of the execution number of the variable is `-1`. But maybe you also want to remove the duplicates of `VARTMP1` and only keep one instance?  Anyway, this behavior should be documented in `mir.mli`.

denismerigoux (changes requested):
Thanks James for this PR! It's looking good overall but miss some polishing. It's okay to not implement all the complex features of #20 if we don't need them :)

denismerigoux:
I think we can close #20 once this is merged.

src/mlang/backend_ir/bir_interface.ml
jemsab: All the praise goes to Keryan for this one!

src/mlang/m_frontend/mast_to_mir.ml
jemsab: `variable_def` rather than `value`, no?

jemsab: As discussed elsewhere, there is limited factoring to be done here, so I'll leave this as is.

denismerigoux (approved):

denismerigoux:
Can I merge ?

jemsab:
With this change everything is good for me if you want to merge

PR #65, opened 2021-09-28
Removed obsolete ppx features

Keryan-dev:
There were exactly *one* remaining occurrence of call to derived functions, that could be replaced with `Stdlib.compare` which should be semantically equivalent.


denismerigoux (approved):
I like PRs that remove code!

PR #75, opened 2021-10-08
Use Github actions instead of custom CI via gitlab

denismerigoux:


.github/workflows/check_correctness.yml
```diff
@@ -0,0 +1,72 @@
...
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      - name: Opam modules cache
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-opam-modules
+        with:
+          # OCaml cache files are stored in `~/.opam` on Linux/macOS
+          path: ~/.opam
```
jemsab, line 32: If you want to stick with `make deps`, local opam switches seem to install in `_opam`  in the project directory rather than `~/.opam`.

From reading the documentation something like this should work : 
```yaml
path: $GITHUB_WORKSPACE/_opam
```

jemsab:
That's a shame, maybe a solution would be using a preconfigured docker image, like what Raphaël was doing on his gitlab runner but with the switch preconfigured as well, that way setup time would be 0 seconds. That will take some time to setup though.

I guess the only problem using a global switch is keeping the compiler versions in sync between the opam file and the github actions yaml.

PR #80, opened 2021-10-13
Fix error raising

Keryan-dev:
Mlang assumed a verification condition could raise several errors at once. But it appears that it always raise exactly one, with sometime an associated variable for what I expect to be reporting purposes.

This may be needed when interfacing with current production practices.

### DIFF
--- a/src/mlang/m_frontend/mparser.mly
+++ b/src/mlang/m_frontend/mparser.mly
@@ -280,9 +280,10 @@ verification:
 } }
 
 verification_condition:
-| IF e = expression THEN ERROR e_names = verification_name+ SEMICOLON { ({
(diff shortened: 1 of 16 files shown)


src/mlang/m_frontend/mparser.mly
```diff
@@ -280,9 +280,10 @@ verification:
 } }
 
 verification_condition:
-| IF e = expression THEN ERROR e_names = verification_name+ SEMICOLON { ({
+| IF e = expression THEN
+  ERROR e_name = verification_name var = output_name? SEMICOLON { ({
```
denismerigoux, line 284: Can you point me to an example in the M code where this happens?

PR #94, opened 2021-11-12
Improve anomaly handling

jemsab:
The DGFiP backend doesn't currently return any information about the errors that occurred during calculation, just that an anomaly occurred. This PR aims to improve on that by returning information for all calculation error categories. 

Relates to #24

### DIFF
--- a/examples/dgfip_c/m_error.h
+++ b/examples/dgfip_c/m_error.h
@@ -14,12 +14,16 @@ typedef struct m_error {
   char minor_code[7];
   char description[81];
   char isisf[2];
-  bool has_occurred;
 } m_error;
 
-int get_occurred_errors(m_error *errors, int size);
+typedef struct m_error_occurrence {
+  m_error *error;
+  bool has_occurred; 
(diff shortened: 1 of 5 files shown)


examples/dgfip_c/m_error.h
```diff
@@ -14,12 +14,17 @@ typedef struct m_error {
   char minor_code[7];
   char description[81];
   char isisf[2];
-  bool has_occurred;
+  char code_information[10];
```
denismerigoux, line 17: Maybe document that the sized alloted for the description and the code information depend on the actual strings present in the M codebase and are susceptible to cause bugs if overflowed?

denismerigoux (approved):

examples/dgfip_c/m_error.h
jemsab: How would you feel about turning them into `const char *` instead ? 
All m_error structs are static in code and therefore don't need a specific size. This also avoids the overflowing problem.

PR #95, opened 2021-11-17
80 columns formatting

Keryan-dev:
Since there is only a few open PRs, I figured it would be the right time to propose this.

In case some people have an annoying time rebasing on top of this, I can provide a more or less automated way to do it painlessly.

(Also a very good way to boost Github stats!)


denismerigoux:
@rmonat no objections?

PR #97, opened 2021-11-22
Add parsing of DGFiP options

ddeclerck:
This PR adds parsing of DGFiP-specific options.
Such options should be specified as `--dgfip-options=o1,...,on`.
The options names are identical to those used by the historical compiler.


denismerigoux (approved):
This is great. Thanks for providing cmdliner doc for these options!

PR #107, opened 2022-01-12
Unify SSA duplicated variables of Mir in Bir

Keryan-dev:
This ought to straighten a bit the re-affectations semantics of Bir programs.

It needs more work to get rid of bugs and performance issues (and even more to smooth the whole thing up), drafting for now.

### DIFF
--- a/src/mlang/mpp_ir/mpp_ir_to_bir.ml
+++ b/src/mlang/mpp_ir/mpp_ir_to_bir.ml
@@ -150,7 +150,9 @@ let translate_m_code (m_program : Mir_interface.full_program)
         | InputVar -> None
         | _ ->
             (* variables used in the context should not be reassigned *)
-            Some (Bir.SAssign (var, vdef), var.Mir.Variable.execution_number.pos)
+            Some
+              ( Bir.SAssign (Bir.var_from_mir var, vdef),
+                var.Mir.Variable.execution_number.pos )
       with Not_found -> None)
     vars
 
@@ -258,7 +260,7 @@ and translate_mpp_stmt (mpp_program : Mpp_ir.mpp_compute list)
         [
           Pos.same_pos_as
             (Bir.SAssign
-               ( new_l,
+               ( Bir.var_from_mir new_l,
                  {
                    var_definition =
                      SimpleVar (translate_mpp_expr m_program ctx expr, pos);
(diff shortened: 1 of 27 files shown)


Keryan-dev:
This now needs #111. [Relevant diff for early review](https://github.com/MLanguage/mlang/pull/107/files/ca9f4a68efb1121a77052de20ec8bee2a119f1b8..5690306b59507e4fc3a0554bdd18bcb13d8020d8)

src/mlang/mpp_ir/mpp_ir_to_bir.ml
```diff
@@ -177,9 +178,44 @@ let translate_m_code (m_program : Mir_interface.full_program)
         let var = Mir.VariableDict.find vid m_program.program.program_vars in
         match vdef.var_definition with
         | InputVar -> None
-        | _ ->
+        | TableVar (i, idx_def) ->
```
rmonat, line 181: Could we use `Mir.map_var_def_var` to factor the code in here?

rmonat:
Out of curiosity, why is the type `variable` described twice in [`mir.ml`](https://github.com/MLanguage/mlang/pull/107/files/ca9f4a68efb1121a77052de20ec8bee2a119f1b8..5690306b59507e4fc3a0554bdd18bcb13d8020d8#diff-200dd6f2bd750e7f5e860abbb468bdd2f7a9cb2b236c36db264403191de232f9R76-R113)?

src/mlang/mpp_ir/mpp_ir_to_bir.ml
Keryan-dev: Indeed. It's done.

Keryan-dev:
> Out of curiosity, why is the type `variable` described twice in [`mir.ml`](https://github.com/MLanguage/mlang/pull/107/files/ca9f4a68efb1121a77052de20ec8bee2a119f1b8..5690306b59507e4fc3a0554bdd18bcb13d8020d8#diff-200dd6f2bd750e7f5e860abbb468bdd2f7a9cb2b236c36db264403191de232f9R76-R113)?

I think it has to do with namespaces used across the compiler such as `Mir.Variable.variable_name` that would not be possible by only aliasing the type. There should be a way to avoid the duplication by reordering some definitions though. I can look into it on a separate patch if you want.

rmonat:
Thanks! I don"t think it's urgent, I was just curious.

PR #112, opened 2022-01-21
Substitute constants by their value in Mast to Mir transformation

### LINKED ISSUE #109: DGFIP backend fails when encountering raw constants
Since #84 which uses `Dgfip_varid` to point at the right place in the TGV instead of on-the-fly index generation, the code generation fails on references to variables that are actually constants.

This should be expected since the compirateur inlines the constants in the code and the TGV contains only actual variables, contrarily to Mlang where the distinction is lost.
The problem resolves itself when using optimizations, that include the behavior of the compirateur in that regard. This is still a bug without it though.

A cheap solution would be to force at least one inlining pass all the time, at least for the DGFIP backend, any opinions on this ?

Keryan-dev:
Closes #109

### DIFF
--- a/src/mlang/m_frontend/mast_to_mir.ml
+++ b/src/mlang/m_frontend/mast_to_mir.ml
@@ -419,7 +419,8 @@ let belongs_to_iliad_app (r : Mast.application Pos.marked list) : bool =
 (** Retrieves variable declaration data. Done in a separate pass because wen
     don't want to deal with sorting the dependencies between files or inside
     files. *)
-let get_variables_decl (p : Mast.program) :
+let get_variables_decl (p : Mast.program)
+    (const_map : float Pos.marked ConstMap.t) :
     var_decl_data Mir.VariableMap.t * Mir.Error.t list * Mir.idmap =
   let vars, idmap, errors, out_list =
     List.fold_left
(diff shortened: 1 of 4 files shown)


src/mlang/m_frontend/mast_to_mir.ml
```diff
@@ -255,10 +256,7 @@ let var_or_int_value (ctx : translating_context) (l : Mast.literal Pos.marked) :
       try
         (* We look up the value of the variable, which has to be const *)
         let name = Mast.get_variable_name v in
-        Mir.VariableMap.find
-          (get_var_from_name ctx.idmap (Pos.same_pos_as name l) ctx.exec_number
-             false)
-          ctx.int_const_values
+        ConstMap.find name ctx.const_map |> Pos.unmark |> int_of_float
```
rmonat, line 259: We could check that there is no loss of precision when calling `int_of_float`

src/mlang/m_frontend/mast_to_mir.ml
```diff
@@ -475,21 +435,36 @@ let get_variables_decl (p : Mast.program)
...
-                      Cli.var_info_print
-                        "Dropping declaration of %s %a because variable was \
-                         previously defined %a"
-                        (Pos.unmark old_var.Mir.Variable.name)
-                        Pos.format_position
-                        (Pos.get_position cvar.Mast.comp_name)
-                        Pos.format_position
-                        (Pos.get_position old_var.Mir.Variable.name);
-                      (vars, idmap, errors, out_list)
+                      match
```
rmonat, line 438: Could we factor some of this code and lines 528-557?

src/mlang/m_frontend/mast_to_mir.ml
Keryan-dev: You mean in case of a non-integer value ? Doesn't the M semantics truncate anyway for array accesses ?

PR #113, opened 2022-01-26
Fully implement java backend

jemsab:
Java backend was only partially implemented, such that if compiling
without optimisations the backend breaks.

This PR aims to rectify that situation allowing the Java backend
to handle unoptimised code.

### DIFF
--- a/examples/java/src/com/mlang/MValue.java
+++ b/examples/java/src/com/mlang/MValue.java
@@ -212,7 +212,13 @@ public class MValue {
     return value.getValue() == 0 ? one : zero;
   }
 
-  static MValue m_multimax(int bound, MValue[] array, int position) {
+  static MValue m_multimax(MValue boundMValue, MValue[] array, int position) {
+    if (boundMValue.isUndefined()) {
+      throw new RuntimeException("Multimax bound undefined!");
+    }
+
+    int bound = (int)boundMValue.getValue();
+
     MValue max = mAdd(array[position], zero);
     for (int i = 0; i <= bound; i++) {
       MValue challenger = mAdd(array[position + i], zero);
@@ -244,7 +250,7 @@ public class MValue {
       return mUndefined;
     } 
      
-    int indexInteger = index.getValue();
+    int indexInteger = (int)index.getValue();
     
     if (indexInteger < 0) {
       return zero;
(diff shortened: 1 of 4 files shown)


examples/java/src/com/mlang/MValue.java
```diff
@@ -239,14 +245,21 @@ static boolean m_is_defined_false(MValue x) {
...
+    if (index.isUndefined()) {
+      return mUndefined;
+    } 
+     
+    int indexInteger = (int)index.getValue();
+    
+    if (indexInteger < 0) {
       return zero;
-    } else if (index >= size) {
+    } else if (indexInteger >= size) {
```
rmonat, line 257: Shouldn't this be `indexInteger > size`?

examples/java/src/com/mlang/MValue.java
```diff
@@ -239,14 +245,21 @@ static boolean m_is_defined_false(MValue x) {
...
   }
 
-  static MValue m_array_index(MValue[] array, int tableStart, int index, int size) {
-    if (index < 0) {
+  static MValue m_array_index(MValue[] array, int tableStart, MValue index, int size) {
+    if (index.isUndefined()) {
+      return mUndefined;
+    } 
+     
+    int indexInteger = (int)index.getValue();
```
rmonat, line 253: The C backend uses `floor` before a cast to int. I would be in favor of doing the same here to be consistent.

src/mlang/backend_compilers/bir_to_java.ml
```diff
@@ -98,14 +92,13 @@ let rec generate_java_expr (e : expression Pos.marked)
...
       (se2, s2)
   | Index (var, e) ->
-      let _, s = generate_java_expr e var_indexes in
-      let index = float_literal_to_int e in
+      let se, s = generate_java_expr e var_indexes in
       let unmarked_var = Pos.unmark var in
       let size = Option.get unmarked_var.Mir.Variable.is_table in
       let se2, s2 =
-        ( Format.asprintf "m_array_index(calculationVariables, %d ,%d, %d)"
+        ( Format.asprintf "m_array_index(calculationVariables, %d ,%s, %d)"
```
rmonat, line 99: Is there a reason to use the name `calculationVariables` rather than `TGV`?

examples/java/src/com/mlang/MValue.java
jemsab: As discussed, I seem to remember doing it without and having problems, however when running the tests today with a strictly greater than, there are no failures. I've therefore made the suggested change.

PR #114, opened 2022-02-02
Fix generic variables name computation

ddeclerck:
This makes generic names computation behave as in the original compiler (fixes potential bugs).

Generic names may be specified using ranges, with the following syntax :
x..y can be used both for integer ranges or **single** character ranges (integers and characters can not be mixed in a range)
x-y is a range whose lower bound is an integer and its upper bound is a **constant**

For instance, one can write such construct :
```
V=somme(i=0..2,4-T,X..Z:Ni)
```

Here, T must be a constant, and X, Y, Z are substituted characters (it does not matter if they also exist as variables or constants).
If T = 6, then we get: N0, N1, N2, N4, N5, N6, NX, NY and NZ.

It is also possible to specify individual substitution strings along with integer ranges.

In any case, the length of the string representation of integers and of the substituted characters/strings must be equal. If necessary, integers are zero-padded on the left so that all lengths become equal. If only integers are present, the reference length is the length of the largest integer's representation.

So, if we write :
```
VARTMP1=somme(i=1..3,XX,YY:NBi);
```

Then we get : NB01, NB02, NB03, NBXX, NBYY

However, if we remove XX and YY, then we get : NB1, NB2, NB3

### DIFF
--- a/src/mlang/m_frontend/mast.ml
+++ b/src/mlang/m_frontend/mast.ml
@@ -98,8 +98,9 @@ type set_value =
 
 (** Values that can be substituted for loop parameters *)
 type set_value_loop =
-  | VarParam of variable_name Pos.marked
-  | IntervalLoop of literal Pos.marked * literal Pos.marked
+  | Single of literal Pos.marked
+  | Range of literal Pos.marked * literal Pos.marked
+  | NumRange of literal Pos.marked * literal Pos.marked
 
 type loop_variable = char Pos.marked * set_value_loop list
 (** A loop variable is the character that should be substituted in variable
--- a/src/mlang/m_frontend/mast_to_mir.mli
+++ b/src/mlang/m_frontend/mast_to_mir.mli
@@ -32,7 +32,7 @@ module ConstMap : Map.S with type key = Mast.variable_name
 module ParamsMap : Map.S with type key = Char.t
 (** Map whose keys are loop parameters *)
 
-type loop_context = loop_param_value ParamsMap.t
+type loop_context = (loop_param_value * int) ParamsMap.t
 (** This is the context when iterating a loop : for each loop parameter, we have
     access to the current value of this loop parameter in this iteration. *)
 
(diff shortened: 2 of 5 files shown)


src/mlang/m_frontend/mast.ml
```diff
@@ -98,8 +98,9 @@ type set_value =
 
 (** Values that can be substituted for loop parameters *)
 type set_value_loop =
-  | VarParam of variable_name Pos.marked
-  | IntervalLoop of literal Pos.marked * literal Pos.marked
+  | Single of literal Pos.marked
+  | Range of literal Pos.marked * literal Pos.marked
+  | NumRange of literal Pos.marked * literal Pos.marked
```
rmonat, line 103: Could we change `NumRange` into `Interval`? I think it would make clearer that it expects to numbers. Contrary to Range where it can either be numbers or names.

src/mlang/m_frontend/mast_to_mir.mli
```diff
@@ -32,7 +32,7 @@ module ConstMap : Map.S with type key = Mast.variable_name
 module ParamsMap : Map.S with type key = Char.t
 (** Map whose keys are loop parameters *)
 
-type loop_context = loop_param_value ParamsMap.t
+type loop_context = (loop_param_value * int) ParamsMap.t
```
rmonat, line 35: The integer corresponds to the expected length of the variable?

rmonat:
Thank you for the detailed explanations and careful reverse-engineering!

src/mlang/m_frontend/mast.ml
ddeclerck: Sure, I'll do that.

src/mlang/m_frontend/mast_to_mir.mli
rmonat: If you could just add a comment I think it would be fine for now.

Issue #117, opened 2022-02-10
MacOS builds failing


jemsab:
MacOS builds are failing because of the errors below, this seems to be because of an update to Cmdliner which should have a fixed version. Cmdliner does not appear in the .opam
```     
  ocamlc src/mlang/.mlang.objs/byte/mlang__Cli.{cmi,cmti} (exit 2)
(cd _build/default && /Users/runner/.opam/4.11.2/bin/ocamlc.opt -w @1..3@5..28@30..39@43@46..47@49..57@61..62-40 -strict-sequence -strict-formats -short-paths -keep-locs -g -bin-annot -I src/mlang/.mlang.objs/byte -I /Users/runner/.opam/4.11.2/lib/ANSITerminal -I /Users/runner/.opam/4.11.2/lib/bytes -I /Users/runner/.opam/4.11.2/lib/cmdliner -I /Users/runner/.opam/4.11.2/lib/dune-build-info -I /Users/runner/.opam/4.11.2/lib/gmp -I /Users/runner/.opam/4.11.2/lib/num -I /Users/runner/.opam/4.11.2/lib/ocaml/threads -I /Users/runner/.opam/4.11.2/lib/ocamlgraph -I /Users/runner/.opam/4.11.2/lib/parmap -I /Users/runner/.opam/4.11.2/lib/re -I /Users/runner/.opam/4.11.2/lib/seq -no-alias-deps -opaque -open Mlang -o src/mlang/.mlang.objs/byte/mlang__Cli.cmi -c -intf src/mlang/utils/cli.mli)
File "src/mlang/utils/cli.mli", line 46, characters 11-29:
46 | val info : Cmdliner.Term.info
                ^^^^^^^^^^^^^^^^^^
Error (alert deprecated): Cmdliner.Term.info
Use Cmd.info instead.
      ocamlc src/mlang/.mlang.objs/byte/mlang__Dgfip_options.{cmi,cmo,cmt} (exit 2)
[...]

PR #118, opened 2022-02-10
Set Cmdliner version to 1.0.4

jemsab:
Following an update to Cmdliner v1.1.0, deprecation warnings now break
the macOS CI build.

Also Cmdliner is a direct dependancy of the project and should therefore be
explicitly mentionned in dune-project and mlang.opam files.


jemsab:
Should close #117

rmonat (approved):

Issue #119, opened 2022-02-10
Update Cmdliner to version 1.1.0


jemsab:
A new release of Cmdliner with breaking changes has been published (https://github.com/dbuenzli/cmdliner/blob/master/CHANGES.md#v110-2022-02-06-la-forclaz-vs), the only dependency is Ocaml > 4.08. We should therefore upgrade the version and make the necessary changes to Mlang's use of this package.

jemsab:
PR merged

PR #120, opened 2022-02-10
Update Cmdliner to v1.1.0

jemsab:
Cmdliner have release a new version 1.1.0 which has breaking changes,
this PR aims to update the version and change all the deprecated
function and module uses.

Closes issue #119


rmonat:
This fail is weird... maybe we should do an `opam update` in the CI (it's suggested in the log)?

PR #129, opened 2022-02-18
Miscellaneous clean-ups

Keryan-dev:
Some things I stumbled upon while working on other things. Might as well have them reviewed outside patches with semantic changes.

### DIFF
--- a/src/mlang/m_ir/mir_typechecker.ml
+++ b/src/mlang/m_ir/mir_typechecker.ml
@@ -16,55 +16,45 @@
 
 open Mir
 
-type ctx = { ctx_is_generic_table : bool }
-
 let is_table_var var =
   match var.Variable.is_table with Some _ -> true | None -> false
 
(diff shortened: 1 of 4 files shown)


src/mlang/m_ir/mir_typechecker.ml
```diff
@@ -75,58 +65,46 @@ let rec typecheck_top_down (ctx : ctx) (e : expression Pos.marked) : ctx =
           ]
 
 and typecheck_func_args (f : func) (pos : Pos.t) :
-    ctx -> Mir.expression Pos.marked list -> ctx =
+    bool -> Mir.expression Pos.marked list -> unit =
   match f with
   | SumFunc | MinFunc | MaxFunc ->
-      fun ctx args ->
+      fun in_generic_table args ->
```
rmonat, line 71: Is there a reason why we keep different `fun in_generic_table args` in the different cases of the `match`?

Keryan-dev: I think not, I factorized all this (among other things).

PR #131, opened 2022-02-22
Distinguish M code in separate chains

Keryan-dev:
This uses tags on M rules definition to identify the subgraph of each "enchaineur" used by INTER files, including corrective ones.

### DIFF
--- a/src/mlang/backend_ir/bir_interpreter.ml
+++ b/src/mlang/backend_ir/bir_interpreter.ml
@@ -479,10 +479,11 @@ module Make (N : Bir_number.NumberInterface) = struct
                 | TableVar _ -> assert false
                 (* should not happen *)
               with Not_found ->
-                Errors.raise_spanned_error
-                  ("Var not found (should not happen): "
-                  ^ Pos.unmark (Bir.var_to_mir var).Mir.Variable.name)
-                  (Pos.get_position e)
+                (* Cli.warning_print
+                 *   "Var %s used before definition, assumed undefined.\n%s"
+                 *   (Pos.unmark var.Variable.name)
+                 *   (Pos.retrieve_loc_text (Pos.get_position e)); *)
+                Undefined
             in
             r
         | GenericTableIndex -> (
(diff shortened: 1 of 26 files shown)


Keryan-dev:
This needs #129 and #107. ([Relevant diff for early reviews](https://github.com/MLanguage/mlang/pull/131/files/cc5482e7ada9378b3e4a3cb119b129d44e6ff8f0..8d467a8088e151025e9d86cc91fddc76c2c65038))

src/mlang/mpp_ir/mpp_ir_to_bir.ml
```diff
@@ -318,9 +327,7 @@ and translate_mpp_stmt (mpp_program : Mpp_ir.mpp_compute list)
                    var_definition =
                      SimpleVar (translate_mpp_expr m_program ctx expr, pos);
                    var_typ = None;
-                   var_io =
-                     (snd (Mir.find_var_definition m_program.program var))
-                       .var_io;
+                   var_io = Mir.Input;
```
rmonat, line 330: Isn't this done by `Bir_interpreter.RegularFloatInterpreter.replace_undefined_with_input_variables` within `wrap_m_code_call`?

Keryan-dev: That does more than just changing the IO type. It removes the `var_definition` entirely so that the M does not override the set value in the M++.
I actually think that is not necessary, nor does it follow the current semantics of INTER files, so we should probably discuss it further at some point.

src/mlang/backend_ir/bir_interpreter.ml
```diff
@@ -479,10 +479,11 @@ module Make (N : Bir_number.NumberInterface) = struct
                 | TableVar _ -> assert false
                 (* should not happen *)
               with Not_found ->
```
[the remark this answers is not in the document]
Keryan-dev: This was needed when variables where duplicated because some where refering to the assigned variable in M code instead of the declaration. It happens in particular with variable defined in corrective rules.
But since variable unification this should not be a problem anymore, I reverted this change.

src/mlang/mpp_ir/mpp_ir_to_bir.ml
rmonat: Okay. Could you please open an issue on this?

rmonat: Thank you! (#136)

PR #138, opened 2022-03-08
[DGFIP] Tests

ddeclerck:
This PR updates the test framework for the DGFiP backend.
Minimal auxiliary files have been written to mimic the DGFiP environment.


examples/dgfip_c/enchain_static.c
```diff
@@ -0,0 +1,211 @@
...
+
+int multimax_def(int nbopd, char *var)
+{
+  for (int i = 0; i < nbopd; i++) {
+    if (var[i] == 1) return 1;
+  }
+  return 0;
+}
+
+double multimax(double nbopd, double *var)
```
rmonat, line 194: `double nbopd`?

examples/dgfip_c/enchain_static.c
```diff
@@ -0,0 +1,211 @@
...
+{
+  double s = 0.0;
+  for (int i = 0; i < (int)nbopd; i++) {
+    if (var[i] >= s) s = var[i];
+  }
+  return s;
+}
+
+int modulo_def(int a, int b)
+{
```
rmonat, line 204: There is no modulo currently supported by the rest of mlang?

examples/dgfip_c/enchain_static.c
```diff
@@ -0,0 +1,211 @@
...
+double floor_g(double a)
+{
+  if (fabs(a) <= LONG_MAX) {
+    return floor(a);
+  } else {
+    return a;
+  }
+}
+
+int multimax_def(int nbopd, char *var)
```
rmonat, line 186: The definition of multimax in other backends seems different from this one (explicit casts of all values in the array).

rmonat (commented):
Thank you very much for the work! I have a few more comments:
- Some incorrectly encoded characters are replaced at some point by Mlang:
```
[WARNING] Replaced char code 160 by space in file /home/raphael/Desktop/mlang-dd/ir-calcul/sources2020m_6_5/errI.m, from 88:31 to 88:105
```
However, the C compiler seem to complain again later:
```
var.c:78:46: warning: illegal character encoding in string literal [-Winvalid-source-encoding]
T_erreur erreur_A090 = { "A090 / LE MONTANT 8<A0>ZT, ZRE ou ZRF EST > AU REVENU NET IMPOSABLE DE LA RUBRIQUE", "090", "00", "N", "A090", 1 };
```
Could you investigate?
- Clang complains that some non-void functions do not return values, both in `ir_tests.c` and `irdata.c`. It also displays a scary warning about a long to double conversion changing values. Could we fix those warnings?
- Could you add `examples/dgfip` within the `clean` target of the root Makefile?
- The C compilation fails for year 2018. Could you check it?

examples/dgfip_c/enchain_static.c
ddeclerck: That's how it's called in the original DGFiP file. I kept the same argument names.

examples/dgfip_c/enchain_static.c
ddeclerck: That function comes from the original DGFiP file. Not sure if it's actually needed.

examples/dgfip_c/enchain_static.c
ddeclerck: That function comes from the original DGFiP file. If other backends differ, maybe we should consider this one as the reference implementation, since it's the one used by the DGFiP.

ddeclerck:
> Thank you very much for the work! I have a few more comments:
> 
> * Some incorrectly encoded characters are replaced at some point by Mlang:
> 
> ```
> [WARNING] Replaced char code 160 by space in file /home/raphael/Desktop/mlang-dd/ir-calcul/sources2020m_6_5/errI.m, from 88:31 to 88:105
> ```
> 
> However, the C compiler seem to complain again later:
> 
> ```
> var.c:78:46: warning: illegal character encoding in string literal [-Winvalid-source-encoding]
> T_erreur erreur_A090 = { "A090 / LE MONTANT 8<A0>ZT, ZRE ou ZRF EST > AU REVENU NET IMPOSABLE DE LA RUBRIQUE", "090", "00", "N", "A090", 1 };
> ```
> 
> Could you investigate?

That's because the substitution occurs late in the backend, while I was assuming it occured during parsing. I'll fix that.

> * Clang complains that some non-void functions do not return values, both in `ir_tests.c` and `irdata.c`. It also displays a scary warning about a long to double conversion changing values. Could we fix those warnings?

Will check.

> * Could you add `examples/dgfip` within the `clean` target of the root Makefile?

Will do.

> * The C compilation fails for year 2018. Could you check it?

Will check.

PR #111, opened 2022-01-21
split mpp functions

jemsab:
Closes #110

PR #143, opened 2022-05-06
Change representation of backend variables

Keryan-dev:
This is bound to change further in the futur, but is good enough for our immediate needs.

We now represent TGV variables as offset of a given TGV in the backend. It will be useful for copies handling.

Most changes in the backend code generation are the direct consequences of the existence of an offset upstream, making `var_indexes` useless. Reviewers shouldn't need to look too hard on those.

### DIFF
--- a/src/mlang/backend_ir/bir.ml
+++ b/src/mlang/backend_ir/bir.ml
@@ -28,8 +28,6 @@ let compare_variable v1 v2 =
   let c = String.compare v1.on_tgv v2.on_tgv in
   if c = 0 then Stdlib.compare v1.offset v2.offset else c
 
-type variable_id = Mir.variable_id
-
 module VariableMap = Map.Make (struct
   type t = variable
 
--- a/src/mlang/backend_ir/bir.ml
+++ b/src/mlang/backend_ir/bir.ml
@@ -36,14 +36,10 @@ module VariableMap = Map.Make (struct
   let compare = compare_variable
 end)
 
-module VariableDict = Dict.Make (struct
-  type t = variable_id
+module VariableSet = Set.Make (struct
+  type t = variable
 
-  type elt = variable
-
-  let key_of_elt v = v.mir_var.Mir.Variable.id
-
-  let compare = compare
+  let compare = compare_variable
 end)
 
 (* unify SSA variables *)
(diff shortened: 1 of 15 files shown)


rmonat:
That seems great! I tried running locally, the Java backend is fine but the DGFiP C backend fails on some tests (`fuzzer_814`, v2020). Are there things that are still missing to support it or is it a bug?

src/mlang/backend_ir/bir.ml
```diff
@@ -251,8 +292,8 @@ let rec remove_empty_conditionals (stmts : stmt list) : stmt list =
          | _ -> stmt :: acc)
        [] stmts)
 
-let rec get_used_variables_ (e : expression Pos.marked) (acc : VariableDict.t) :
-    VariableDict.t =
+let rec get_used_variables_ (e : expression Pos.marked) (acc : VariableSet.t) :
```
rmonat, line 295: (Nitpicking) I think we have a fold and a map over expressions, so this piece of code could be refactored.

src/mlang/backend_ir/bir.ml
```diff
@@ -18,30 +18,72 @@ type rule_id = Mir.rule_id
...
+end)
 
-let var_to_mir v = v
+module NameMap = Map.Make (String)
+
+type offset_alloc = { mutable name_map : int NameMap.t; mutable size : int }
+
+(* Mutable state hidden away in the signature. Used for black-magicaly
+   transition variable representations from SSA duplications to offsets of TGV.
+   An issue though: this disregards tetantives to reduce the size of the TGV
```
rmonat, line 52: I guess we are kind of stuck with keeping the TGV expected by the clients. However we can still get rid of the computations?

Keryan-dev: Eventually ? Optimisations are disable for most backends right now (since #111), but we could rework an inlining/specialisation decision at some point, would the need arise.

Also, the TGV representation in most backend is different from the current DGFIP workflow nonetheless, but it shouldn't matter as long as clients use the provided interface (currently, mostly `desc.c`).

Keryan-dev:
The DGFIP backend won't work until we can properly generate code for the primitive calculation (and not just chain function as it is the case now), i.e. sprint 2. This patch works toward that, others will come along.

src/mlang/backend_ir/bir.ml
Keryan-dev: Done

PR #146, opened 2022-05-18
Relax M table access to variables

Keryan-dev:
This gets Mlang closer to the expressivity of the legacy compiler by allowing access to table variables through dynamic indexing.
It also removes the hack with the implicit loop variable `X`.

I feel this is not entirely satisfactory as it introduce writes to specific cells of array, whereas Mlang tried to represent table writes as whole array modifications until now. We should smooth up the way table variables are handled, I'm not sure how.

This also introduces potential out of bound accesses unobservable before runtime. I implemented proper boundary checks in the interpreter only.

### DIFF
--- a/src/mlang/backend_ir/bir_interpreter.ml
+++ b/src/mlang/backend_ir/bir_interpreter.ml
@@ -64,7 +64,6 @@ module type S = sig
   type ctx = {
     ctx_local_vars : value Pos.marked Mir.LocalVariableMap.t;
     ctx_vars : var_value Bir.VariableMap.t;
-    ctx_generic_index : int option;
   }
 
   val empty_ctx : ctx
@@ -160,14 +159,12 @@ module Make (N : Bir_number.NumberInterface) = struct
   type ctx = {
     ctx_local_vars : value Pos.marked Mir.LocalVariableMap.t;
     ctx_vars : var_value Bir.VariableMap.t;
-    ctx_generic_index : int option;
   }
 
   let empty_ctx : ctx =
     {
       ctx_local_vars = Mir.LocalVariableMap.empty;
       ctx_vars = Bir.VariableMap.empty;
-      ctx_generic_index = None;
     }
 
   let literal_to_value (l : Mir.literal) : value =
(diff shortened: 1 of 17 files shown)


src/mlang/backend_ir/bir_interpreter.ml
```diff
@@ -649,20 +641,52 @@ module Make (N : Bir_number.NumberInterface) = struct
...
+                  | SimpleVar n -> n
+                  | TableVar _ -> assert false
+                  (* should not happen *)
+                  | exception Not_found ->
+                      raise
+                        (RuntimeError
+                           ( ErrorValue
+                               ( "no value found for dynamic index",
+                                 Pos.get_position e ),
+                             ctx ))
```
rmonat, line 664: Could you confirm you have checked that the original compiler does not check out of bound accesses?

rmonat:
> I feel this is not entirely satisfactory as it introduce writes to specific cells of array, whereas Mlang tried to represent table writes as whole array modifications until now. We should smooth up the way table variables are handled, I'm not sure how.

We'll need to discuss that. I think we had whole modifications because we thought the original compiler was working this way.

rmonat:
I am encountering some issues with the DGFiP C backend:
- with `OPTIMIZE=0`, one test fails: `Testing file: fuzzer_2505.m_test Expected value for ICREREVET : 0.0000, computed 4.0000!`
- with `OPTIMIZE=1`, I have a compilation failure, `./ir_tests.h:4:2: error: unterminated conditional directive
 #ifndef IR_HEADER_`. Maybe the effect of `generate_footer` is not flushed correctly before the file is closed?

src/mlang/backend_ir/bir_interpreter.ml
Keryan-dev: It does check the lower bound (0) but not the upper one. The code below is stricter than this, which I think is fair enough.

Here it checks if the variable index is valuated and crash if not, whereas the old compiler just discards the assign silently, as he would do for negative index.

PR #148, opened 2022-05-25
OCaml backend

mdurero:


src/mlang/backend_compilers/bir_to_ocaml.ml
```diff
@@ -0,0 +1,383 @@
...
+  let pp_print_line fmt variable : unit =
+    match get_position_and_alias variable with
+    | Some (position, alias) ->
+        Format.fprintf fmt
+          "let tgv_positions = TgvPositionMap.add \"%s\" %d tgv_positions in"
+          alias position
+    | None -> ()
+  in
+  let pp_print_position_map fmt input_vars =
+    Format.pp_print_list pp_print_line fmt input_vars
```
Keryan-dev, line 342: Using `pp_print_list` with default argument `~pp_sep` set to a cut will leave blank lines in a vertical box, with the expected displeasing result in the generated file.

src/mlang/backend_compilers/bir_to_ocaml.ml
```diff
@@ -0,0 +1,383 @@
...
+  let output_vars =
+    List.map fst (Bir.VariableMap.bindings function_spec.func_outputs)
+  in
+  let name_and_pos_list var_list =
+    List.map
+      (fun var -> (get_var_pos var, Pos.unmark var.mir_var.name))
+      var_list
+  in
+  let print_line fmt (position, name) =
+    Format.fprintf fmt "{alias = \"%s\" ; value = (Array.get tgv %d).value} ::"
```
Keryan-dev, line 315: Please avoid generating calculation for static values. Here the list could be built as a single block rather than chaining `::`. This actually causes native compilation to stack overflow (`make runx`).

MdrditArthur: Would something along the lines of

```suggestion  
    let print_line fmt (position, name) =
      Format.fprintf fmt "{alias = \"%s\" ; value = (Array.get tgv %d).value}"
        name position
    in 
    let pp_print_output_get fmt output_vars =
      Format.pp_print_list ~pp_sep:pp_list_element_separator print_line fmt (name_and_pos_list output_vars)
    in
    Format.fprintf oc "let output (tgv : m_array) : output_list =@,[%a]@," pp_print_output_get output_vars
```
with a pp_list_element_separator inserting a simple " ; " between each alias&value pair do the trick?

Or do I misunderstand your suggestion of building the list as a single block?

Keryan-dev: Yes that's what I meant. Although thinking about it some more, I'm not sure that will do it. Maybe it's safer to implement it dynamically (i.e. having the iterator in the generated ocaml instead of the code generator)

MdrditArthur: It doesn't solve the stack overflow in any case. I ran into the stack problem quickly, right after commit 602e8a88165b8d361f3ff02ef086f4e6d0ef63e7 in June.

src/mlang/backend_compilers/bir_to_ocaml.ml
MdrditArthur: Easily solved by using an empty separator and moving the break hint up in the « Some » member of the match, avoiding to insert a cut resulting in a blank line in the « None » case.

MdrditArthur:
Native compilation of the generated code file causes a stack overflow, due to multiple huge dynamic data structures for entries and outputs. As a first approach, we could modify generated code to build the corresponding static arrays as the first step, then apply relevant transformation to compute the dynamic parts and convert to the target data structure.

examples/ocaml/test_harness.ml
```diff
@@ -0,0 +1,255 @@
...
+  | head :: tail when String.starts_with ~prefix:"#" head ->
+      { block_name = head; block_data = parse_codes tail }
+  | _ -> failwith "Data block has not the expected structure."
+
+let rec split_blocks (line_list : string list)
+    (line_list_list : string list list) : string list list =
+  match line_list with
+  | [] -> List.rev (List.map List.rev line_list_list)
+  | line :: tail ->
+      if String.starts_with ~prefix:"#" line then
```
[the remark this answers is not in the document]
MdrditArthur: Should be fixed by using the parser from the test_framework in the main mlang codebase, instead of doing the same job badly and buggyly.

src/mlang/backend_compilers/bir_to_ocaml.ml
MdrditArthur: Solved, including overflow, by keeping the built array static (array of name and position), then applying it a transformation to get the value. See conversation about stack overflow below.

examples/ocaml/test_harness.ml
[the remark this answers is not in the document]
MdrditArthur: The backend is now usable with compiler 4.11.2.

I have copied selected parts of Mlang internal parser and test parser in order to obtain a independent parser for DGFiP test format. It is incomplete (as it was in Mlang test), but I think it would be a good idea to extend it.

examples/ocaml/mvalue.ml
```diff
@@ -0,0 +1,164 @@
...
+  if x.undefined then m_undef
+  else
+    {
+      undefined = false;
+      value =
+        (if x.value < 0.0 then ceil (x.value -. 0.50005)
+        else floor (x.value +. 0.50005));
+    }
+
+let m_null = m_not
```
Keryan-dev, line 138: Habile

examples/ocaml/mvalue.ml
```diff
@@ -0,0 +1,164 @@
...
+      let new_max =
+        m_max reference (get_position_value current_index)
+      in
+      if current_index = max_index then new_max
+      else multimax variable_array (current_index + 1) max_index new_max
+    in
+    if bound >= 1 then
+      multimax variable_array (position + 1) (position + bound)
+        (get_position_value position)
+    else get_position_value position
```
Keryan-dev, line 164: This `if` would not be needed if the exit condition above was more robust.
Hint: you can write the whole thing more concisely with `Array.sub` and `Array.fold`.

I don't doubt the current implementation is correct, mind.

examples/ocaml/mvalue.ml
```diff
@@ -0,0 +1,164 @@
...
+    let bound = int_of_float bound_variable.value in
+    let get_position_value position =
+      Array.get variable_array position
+    in
+    let rec multimax variable_array current_index max_index reference =
+      let new_max =
+        m_max reference (get_position_value current_index)
+      in
+      if current_index = max_index then new_max
+      else multimax variable_array (current_index + 1) max_index new_max
```
Keryan-dev, line 159: This is fine until you call `multimax` with a start index greater than the max index.

Keryan-dev:
I'd like to have #184 merged to have a more clean diff on this PR for reviewing. @denismerigoux ?

As an early thought: parts of it rely a lot on makefiles whereas we could use dune for a less tedious build system.
On a related note, we could have the test parser exposed as a library for both the compiler and the harness, avoiding code duplication (and possibly more extensive features and tooling in the future).

But it works as it is, so we can just keep that in mind for later patches.

Keryan-dev:
Also could you add this backend to the CI ?

MdrditArthur:
Making the test parser an independent library and fleshing it out to support all our test files is precisely my plan, if I'm not asked to prioritise something else.
I would like to go farther but at the moment I don't clearly see the external uses we could have for this tool (for instance a test-validator to automate syntax check when test files are delivered by our business owners).

As for adding the backend to the CI, I have this in my Git stash since I don't know when, applying it when I need it. At that time I was squeamish about modifying Mlang code base so I never committed the changes on the main Makefile. Just have to check if I use custom test files because I had to tailored some of them to test M errors when I added them.

denismerigoux:
#184 merged

src/mlang/backend_compilers/bir_to_ocaml.ml
```diff
@@ -0,0 +1,387 @@
...
+      (format_tgv_set tgv_expression)
+      position
+  in
+  match vdata.var_definition with
+  | SimpleVar e -> generate_one_var (get_var_pos variable) oc e
+  | TableVar (_, IndexTable es) ->
+      let bindings_list = Mir.IndexMap.bindings es in
+      Format.pp_print_list ~pp_sep:pp_statement_separator
+        (fun fmt (i, v) ->
+          generate_one_var (get_var_pos variable |> ( + ) i) fmt v)
```
Keryan-dev, line 158: Well, it certainly is one way to write this 😀

src/mlang/backend_compilers/bir_to_ocaml.ml
```diff
@@ -0,0 +1,387 @@
...
+      let pos = Pos.get_position stmt in
+      let fname =
+        String.map
+          (fun c -> if c = '.' then '_' else c)
+          (Filename.basename (Pos.get_file pos))
+      in
+      let cond_name =
+        Format.asprintf "cond_%s_%d_%d_%d_%d" fname (Pos.get_start_line pos)
+          (Pos.get_start_column pos) (Pos.get_end_line pos)
+          (Pos.get_end_column pos)
```
Keryan-dev, line 238: Is this for debugging purposes ? Seems a bit overkill for a shadowable name.

src/mlang/backend_compilers/bir_to_ocaml.ml
```diff
@@ -0,0 +1,387 @@
...
+     let entry_positions_map = @,\
+     Array.fold_left (fun pos_map (alias, position) -> TgvPositionMap.add \
+     alias position pos_map)@,\
+     TgvPositionMap.empty entry_positions_array in@,\
+     let init_tgv_var (entry_var : revenue_code) : unit =@,\
+     Array.set tgv (TgvPositionMap.find entry_var.alias entry_positions_map)@,\
+     {undefined = false ; value = entry_var.value} in@,\
+     List.iter init_tgv_var entry_list" pp_print_position_map input_vars
+(* Prévoir les cas : variable manquante, variable en trop dans entry_list,
+   variable définie n fois*)
```
Keryan-dev, line 359: Is it done ?

Keryan-dev (approved):
Some minor comments, but it seems good to merge.

There are a few points of improvement to make which I already discussed a while back with @MdrditArthur. But it's not important enough to delay the merge any further IMO.

examples/ocaml/mvalue.ml
Keryan-dev: That's my point above: you shouldn't need the `>= 1` check with a stronger exit condition. But all this won't matter with a fold anyway.

MdrditArthur: I agree. Well I don't know how, but I manage to break it with the fold…

src/mlang/backend_compilers/bir_to_ocaml.ml
MdrditArthur: I would say: "probably close to how it was done in other backends" and the only use I picture for it in the foreseeable future is helping generated code readability.

src/mlang/backend_compilers/bir_to_ocaml.ml
MdrditArthur: Good catch. No it isn't. Seems like all tests are kindly providing well built entry lists. Not so obvious a check to add, though.

Keryan-dev: Don't bother, just put a comment stating we don't check those cases.

src/mlang/backend_compilers/bir_to_dgfip_c.ml
```diff
@@ -284,8 +284,8 @@ let rec generate_c_expr (e : expression Pos.marked)
         Dand
           ( Dand
               ( Dvar (Local idx_var, Def),
-                Dbinop
-                  ("<=", Dvar (Local idx_var, Val), Dlit (float_of_int size)) ),
+                Dbinop ("<", Dvar (Local idx_var, Val), Dlit (float_of_int size))
+              ),
             Daccess (Pos.unmark var, Def, idx_var) )
```
MdrditArthur, line 289: This (and the Java equivalent) has technically no link with OCaml backend and should be cherry-picked in its own PR

MdrditArthur: Done. Add dependency to PR#196 to this one (administrative dependency, so to speak)

PR #158, opened 2022-06-16
Generate rules in separate files corresponding to sources in DGFIP backend

Keryan-dev:
With this Mlang should generate the same files as the original compiler (modulo verifications for now).
Only in the DGFIP backend, I'm not sure this is something we want in the others.

I didn't notice a lot of change in compilation performances, though.

### DIFF
--- a/src/mlang/backend_ir/bir.ml
+++ b/src/mlang/backend_ir/bir.ml
@@ -177,7 +177,9 @@ let squish_statements (program : program) (threshold : int)
     let id = Mir.fresh_rule_id () in
     {
       rule_id = id;
-      rule_name = (rule_suffix ^ string_of_int id, Pos.no_pos);
+      rule_name =
+        ( rule_suffix ^ string_of_int (Mir.num_of_rule_or_verif_id id),
+          Pos.no_pos );
       rule_code = Rule (List.rev stmts);
     }
   in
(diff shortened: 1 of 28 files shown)


src/mlang/backend_ir/bir.ml
```diff
@@ -95,17 +95,26 @@ type variable_data = variable Mir.variable_data_
...
+}
 
 and stmt = stmt_kind Pos.marked
 
 and stmt_kind =
   | SAssign of variable * variable_data
   | SConditional of expression * stmt list * stmt list
   | SVerif of condition_data
-  | SRuleCall of rule_id
+  | SRuleCall of rov_id
```
rmonat, line 112: I'm a bit confused here: is `rov_id` always an identifier for a `Rule`, or can it be a `Verif` too?

rmonat (commented):
Thank you very much Keryan!

src/mlang/backend_ir/bir.ml
Keryan-dev: Ah, that's a miss. Will patch

Keryan-dev:
> Also, are you able to `make test_dgfip_c_backend`? Could we now add these backend tests in the CI?

Not without proper primitive calculation. Although maybe it's time to adapt the .mpp of other backend.

PR #162, opened 2022-06-23
Handle EBCDIC encoding for variable table sort

jemsab:
Certain clients use systems with EBCDIC encoding for strings, the
variable tables need to be ordered such as they are compatible with
EBCDIC encoding instead of ASCII. This PR performs the varable sort
based on EBCDIC if dgfip option -b1 is present

### DIFF
--- a/src/mlang/backend_compilers/dgfip_gen_files.ml
+++ b/src/mlang/backend_compilers/dgfip_gen_files.ml
@@ -17,6 +17,151 @@
 module StringSet = Set.Make (String)
 module StringMap = Map.Make (String)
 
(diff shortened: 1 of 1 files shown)


src/mlang/backend_compilers/dgfip_gen_files.ml
```diff
@@ -17,6 +17,151 @@
 module StringSet = Set.Make (String)
 module StringMap = Map.Make (String)
 
+let ascii_to_ebcdic =
```
rmonat, line 20: Is there a simple way to format $x$ numbers per line?

PR #165, opened 2022-06-24
Refactor DGFIP backend

Keryan-dev:
The backend is currently a mess of strings passed around the code because of #84 and the way decoupling of definition and value was made. This aims to rectify that by introducing an intermediary AST to properly transform internal expressions into easily printable decoupled ones, and possibly be smarter about it.


src/mlang/mpp_ir/mpp_ir_to_bir.ml
```diff
@@ -254,10 +254,12 @@ let generate_verif_call (m_program : Mir_interface.full_program)
...
-        let res = Mast.compare_error_type (fst cond1.Mir.cond_error).typ
-            (fst cond2.Mir.cond_error).typ in
-        if res <> 0 then res
-        else Stdlib.compare v1.Mir.Variable.id v2.Mir.Variable.id)
+           let res =
+             Mast.compare_error_type (fst cond1.Mir.cond_error).typ
+               (fst cond2.Mir.cond_error).typ
+           in
+           if res <> 0 then res
+           else Stdlib.compare v1.Mir.Variable.id v2.Mir.Variable.id)
```
Keryan-dev, line 262: This is just formatting that slipped of past PRs. Ignore.

Keryan-dev:
Given the coming changes to backend passes, I'm not sure it's worth spending time on this now.

Beside, it's only useful for M expressions (i.e. mostly RHS of affectations).

PR #166, opened 2022-06-28
CI test for DGFIP backend

Keryan-dev:
From an ML rewrite of the primitive calculation, as a temporary solution until new M++.

It's a bit slow though, maybe we should disable the other C backend to keep the CI reasonably fast ?


examples/dgfip_c/ml_primitif/main.ml
```diff
@@ -0,0 +1,94 @@
...
+           code montant montant')
+    ) expected_tgv;
+  !result
+
+let run_test test_file =
+  Printf.printf "Testing %s...\n%!" test_file;
+
+  let tgv, res_prim  = read_test test_file in
+
+  (* Normalement fait par règle 1 *)
```
rmonat, line 48: Should be fixed by #169?

Keryan-dev: I will try to remove that once #169 is merged.

PR #168, opened 2022-06-29
[DGFiP] Fix auxiliary file generation when using EBCDIC

ddeclerck:


src/mlang/backend_compilers/dgfip_gen_files.ml
```diff
@@ -530,9 +566,9 @@ let gen_table fmt (flags : Dgfip_options.flags) vars req_type opt =
 
   Format.fprintf fmt "};\n"
 
-let gen_desc fmt vars ~alias_only =
+let gen_desc fmt vars ~alias_only is_ebcdic =
   (* EBCDIC order not important as no binary search is done on descs*)
```
jemsab, line 570: This comment should probably be removed

PR #173, opened 2022-07-20
[DGFiP] Fix verifications

ddeclerck:


src/mlang/mpp_ir/mpp_ir_to_bir.ml
```diff
@@ -109,7 +109,7 @@ let var_filter_compatible_subtypes (subtypes : Mir.variable_subtype list)
           List.exists
             (fun st ->
               match (st : Mir.variable_subtype) with
-              | Base | GivenBack | Computed -> true
```
Keryan-dev, line 112: Out of the loop here, could you explain this ?

ddeclerck: So called "given back" (restituee) variables are not really a variable subtype by themselves, rather an attribute. Furthermore, it does not only apply to "computed" variables : there are almost 1000 "input" (saisie) variables that are also "given back" (because, you know, WTF...). Anyways, removing "GivenBack" here fixes a bug (mlang-generated verification chainings called more individual verifications than compir-generated verifications chanings).

src/mlang/mpp_ir/mpp_ir_to_bir.ml
```diff
@@ -263,7 +263,7 @@ let generate_verif_call (m_program : Mir_interface.full_program)
       Pos.map_under_mark
         (fun verif_id -> Bir.SRovCall verif_id)
         verif.Mir.cond_number)
-    verifs
+    verifs |> List.rev
```
ddeclerck, line 266: And in case you wonder, this fixes the call order of individual verifications to match the order produced by compir (there is probably a better way to fix this).

src/mlang/backend_compilers/bir_to_dgfip_c.ml
```diff
@@ -581,9 +581,9 @@ let generate_mpp_function (program : Bir.program)
     mppf_stmts
     (if ret_type then
      {|#ifdef FLG_MULTITHREAD
-      return irdata->tas_discord;
+      return irdata->discords;
 #else
-      return tas_discord;
+      return discords;
```
ddeclerck, line 586: And for this one, we weren't returning the correct discord array (tas_discord is used internally to reuse previously allocated discord records, so it points to the set of errors of the previous test case)

PR #178, opened 2022-08-31
Remove C and Python backends

Keryan-dev:
These backends are more or less obsolete and tedious to maintain. Removing them unburden development and CI.

Maybe we should tag master commit before this PR for bookkeeping ?


denismerigoux:
I've added a tag.

denismerigoux (approved):

PR #180, opened 2022-09-01
Multimax semantics fix

### LINKED ISSUE #142: Semantics of multimax differs from the original one
Cf. https://github.com/MLanguage/mlang/blob/51a2aaa37f7d1bd6fa757345bd647c0018a156c6/examples/dgfip_c/enchain_static.c#L186

Small changes wrt to when multimax returns a defined result.

Keryan-dev:
Closes #142 

`multimax` was though to be always defined, but it turns out it's only if any of its component is too.

Patched on top of #165 for simplicity, and silently dependent of #178.

### DIFF
--- a/src/mlang/backend_ir/bir_interpreter.ml
+++ b/src/mlang/backend_ir/bir_interpreter.ml
@@ -571,8 +571,13 @@ module Make (N : Bir_number.NumberInterface) = struct
             in
             let maxi = ref (access_index 0) in
             for i = 0 to Int64.to_int up do
-              (* Fragile: rely on polymorphic compare where None < Some _ *)
-              maxi := max !maxi (access_index i)
+              match access_index i with
+              | None -> ()
+              | Some n ->
+                  maxi :=
+                    Option.fold ~none:(Some n)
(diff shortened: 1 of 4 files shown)


src/mlang/backend_ir/bir_interpreter.ml
```diff
@@ -571,6 +571,7 @@ module Make (N : Bir_number.NumberInterface) = struct
             in
             let maxi = ref (access_index 0) in
             for i = 0 to Int64.to_int up do
+              (* Fragile: rely on polymorphic compare where None < Some _ *)
```
denismerigoux, line 574: Can you explicit this comparison with a match? This is the place where the reference semantics is so we want it to be very explicit.

denismerigoux (changes requested):
So the fuzzed tests did not catch this semantic mismatch?

Keryan-dev:
The fuzzing wouldn't be able to catch that, there is a single use of multimax in 2020 whose result is guarded by a casting operation.

denismerigoux (approved):

PR #184, opened 2022-09-12
Allow for selection of diversly contextualized main functions in Bir

MdrditArthur:
Bir_interface currently overwrites silently MPP-declared main function by a "context_function", adding reset of input variables and valuation of constants and conditional statements to verify according to m_spec file used. If no m_spec is used, a reset of the whole input table is added.

This results in unwanted behaviour for some backends and obfuscates which part of the statements originates from MPP files or M_spec files.

Target for this PR as for now is : 
- Bir module public interface should provide access to all three of bare main_function, main_function with context from a m_spec (context_function) and context_function providing a reset of the input table (needed for the Mlang built-in interpreter)
- Bir.program function map should stay untouched by Bir_interface, reflecting the computation as described in M and MPP, that's why statements for context_function and Co. should be stored in specific fields of the program record.

### DIFF
--- a/src/mlang/backend_ir/bir.mli
+++ b/src/mlang/backend_ir/bir.mli
@@ -61,6 +61,7 @@ type program = {
   mpp_functions : mpp_function FunctionMap.t;
   rules_and_verifs : rule_or_verif ROVMap.t;
   main_function : function_name;
+  context_function : function_name;
   idmap : Mir.idmap;
   mir_program : Mir.program;
   outputs : unit VariableMap.t;
@@ -83,6 +84,8 @@ val set_from_mir_dict : tgv_id -> Mir.VariableDict.t -> VariableSet.t
 val rule_or_verif_as_statements : rule_or_verif -> stmt list
 
 val main_statements : program -> stmt list
+
+val main_statements_no_context : program -> stmt list
 
 val get_all_statements : program -> stmt list
 
--- a/src/mlang/optimizing_ir/bir_to_oir.ml
+++ b/src/mlang/optimizing_ir/bir_to_oir.ml
@@ -203,6 +203,7 @@ and re_translate_block (block_id : Oir.block_id)
       let stmts = List.rev stmts in
       (next_block_id, stmts, rules)
 
+(*WARNING : OIR is not tested, but change in Bir interface to main/context functions has probably broken it.*)
 let oir_program_to_bir (p : Oir.program) : Bir.program =
   let statements, rules_and_verifs =
     re_translate_blocks_until p.entry_block p.blocks Bir.ROVMap.empty None
@@ -222,4 +223,5 @@ let oir_program_to_bir (p : Oir.program) : Bir.program =
     mir_program = p.mir_program;
     outputs = p.outputs;
     main_function = p.main_function;
+    context_function = p.main_function;
   }
--- a/src/mlang/test_framework/test_interpreter.ml
+++ b/src/mlang/test_framework/test_interpreter.ml
@@ -159,7 +159,7 @@ let add_test_conds_to_combined_program (p : Bir.program)
       conds []
   in
   let mpp_functions =
-    Bir.FunctionMap.add p.Bir.main_function
+    Bir.FunctionMap.add p.Bir.context_function
       Bir.{ mppf_stmts = new_stmts @ conditions_stmts; mppf_is_verif = false }
       p.mpp_functions
   in
--- a/src/mlang/backend_ir/bir.ml
+++ b/src/mlang/backend_ir/bir.ml
@@ -127,12 +127,19 @@ type program = {
   mpp_functions : mpp_function FunctionMap.t;
   rules_and_verifs : rule_or_verif ROVMap.t;
   main_function : function_name;
+  context_function : function_name;
   idmap : Mir.idmap;
   mir_program : Mir.program;
   outputs : unit VariableMap.t;
 }
 
 let main_statements (p : program) : stmt list =
+  try (FunctionMap.find p.context_function p.mpp_functions).mppf_stmts
+  with Not_found ->
+    Errors.raise_error
+      "Unable to find contextualized main function of Bir program"
+
+let main_statements_no_context (p : program) : stmt list =
   try (FunctionMap.find p.main_function p.mpp_functions).mppf_stmts
   with Not_found ->
     Errors.raise_error "Unable to find main function of Bir program"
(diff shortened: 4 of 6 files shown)


src/mlang/backend_ir/bir.mli
```diff
@@ -61,6 +61,8 @@ type program = {
   mpp_functions : mpp_function FunctionMap.t;
   rules_and_verifs : rule_or_verif ROVMap.t;
   main_function : function_name;
+  context_function : mpp_function;
+  context_with_reset_function : mpp_function;
```
Keryan-dev, line 65: I'd rather have those out of the function map, since it corresponds to nothing in the mpp. Maybe a option would be to put the statements lists out in the record (or a context subrecord) with names such as e.g. `undefined_inputs_init` `constant_inputs_init` and `adhoc_spec_tests`. Then make sure the proper statements list are build by function below and that the rest of the code refer to those functions instead of tapping into the program record.

Keryan-dev: I wouldn't worry for OIR as long as you input the same date as before (and even then...)

MdrditArthur: The point with OIR is: it transforms Bir.program to Oir.program and back, this could be solved easily by adding relevant information to Oir.program. 
But I don't know yet if the context part of the statements were included in the optimisation. I think so, because it was added to the function map and indistinguishable from the original statements of the program. If so, my work will have changed the optimisation behaviour.

Keryan-dev: Optimisations already lose information by burying it in its output code. If you give the same input (with context IIRC), then you'd have the same output, losing context information in the process, sure, but its output should not change.

src/mlang/test_framework/test_interpreter.ml
```diff
@@ -120,51 +120,6 @@ let to_MIR_function_and_inputs (program : Bir.program) (t : test_file)
   ( { func_variable_inputs; func_constant_inputs; func_outputs; func_conds },
     input_file )
 
-let add_test_conds_to_combined_program (p : Bir.program)
-    (conds : Bir.condition_data Bir.VariableMap.t) : Bir.program =
-  (* because evaluate_program redefines everything each time, we have to make
-     sure that the redefinitions of our constant inputs are removed from the
-     main list of statements *)
```
MdrditArthur, line 127: I suggest we remove this function (only used in this file) as it looks like it has no more use. When it's called, the conditions context statement have already been added, and the context variables initialisation is done with SimpleVar type so no InputVar needs to be filtered and replaced.

src/mlang/backend_ir/bir.mli
MdrditArthur: If I keep this behaviour, the OIR will effectively bury the potential context in the mpp_functions map of the Bir.program. I feel like this would be an unwanted behaviour as part of the point of having the context isolated in its own field is to keep the ability to choose in the backends if we want to use a contextualized main function or not.

So my position would be that OIR should keep either keep intact the context statements (but then they won't be optimized), or the choice to have them should be made before OIR call, as far as I can tell it's not trivial with the current Driver.

I'm inclined to modify OIR.program to pass context information transparently as the context is not the most useful in the cases we want to optimise the generated code (like C for batch processing).

Keryan-dev: Let it drown, without the context some passes might have the wrong behaviour, and you don't want to put your nose in there for this PR. 

I agree this is unsatisfactory but it's not worth spending much time on this for now. We'll discuss this in more details later.

MdrditArthur: > Let it drown,

Well, now we can't, because optimisation made it back to master and so this PR breaks it as is, so I'm trying to resolve it while staying close of the old behaviour.

src/mlang/optimizing_ir/bir_to_oir.ml
```diff
@@ -177,6 +178,10 @@ and re_translate_block (block_id : Oir.block_id)
 let cfg_to_bir_stmts (cfg : Oir.cfg) : Bir.stmt list =
   re_translate_blocks_until cfg.entry_block cfg.blocks None
 
+(*WARNING : OIR is not tested, but changes in Bir interface to "context" of the
+  program (Bir_interface.adapt_program_to_function) could have broken it. In any
+  cases : its behavior is modified as the context is no more included in the
+  optimisations.*)
```
Keryan-dev, line 184: Outdated comment

MdrditArthur: Of course. Isn't comment function to become outdated and treacherous ? I'll delete this ASAP.

src/mlang/test_framework/test_interpreter.ml
denismerigoux: Ok!

src/mlang/backend_ir/bir.mli
```diff
@@ -57,10 +57,17 @@ type mpp_function = { mppf_stmts : stmt list; mppf_is_verif : bool }
 
 module FunctionMap : Map.S with type key = function_name
 
+type program_context = {
```
denismerigoux, line 60: I'm ok with the change but you should put the rationale about why we need this context as a documentation comment here (you can start from your PR description which is quite informative).

src/mlang/backend_ir/bir_interpreter.ml
```diff
@@ -752,7 +752,9 @@ module Make (N : Bir_number.NumberInterface) = struct
       (code_loc_start_value : int) : ctx =
     try
       let ctx =
-        evaluate_stmts p ctx (Bir.main_statements p) [] code_loc_start_value
+        evaluate_stmts p ctx
+          (Bir.main_statements_with_reset p)
```
denismerigoux, line 756: Here a little comment about why we reset would be good

denismerigoux (approved):
Congrats @MdrditArthur ! The code is fine but lacks some comments. You can put comments with the `(* .. *)` syntax, or `(** .. *)` for top-level comments that will be collected by `ocamldoc` to generate https://mlanguage.github.io/mlang/mlang/index.html

src/mlang/backend_ir/bir.mli
MdrditArthur: Documentation comment proposed in commit [b8e0f56](https://github.com/MLanguage/mlang/pull/184/commits/b8e0f56a440017d640a2a4646980fb974bbaa939).

PR #186, opened 2022-09-16
Rehabilitate partial evaluation as a per-function optimisation

Keryan-dev:
This allow a few simplification in the generated code.

Dead code and inlining have gone through the same treatment but have been disabled completely as it would be difficult to test their correctness (not that I think it can be correct in this context).

### DIFF
--- a/src/mlang/backend_compilers/bir_to_dgfip_c.ml
+++ b/src/mlang/backend_compilers/bir_to_dgfip_c.ml
@@ -97,7 +97,7 @@ let rec format_dexpr (dgfip_flags : Dgfip_options.flags)
   match de with
   | Done -> Format.fprintf fmt "1"
   | Dzero -> Format.fprintf fmt "0"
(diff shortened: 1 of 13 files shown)


src/mlang/backend_compilers/bir_to_dgfip_c.ml
```diff
@@ -97,7 +97,7 @@ let rec format_dexpr (dgfip_flags : Dgfip_options.flags)
   match de with
   | Done -> Format.fprintf fmt "1"
   | Dzero -> Format.fprintf fmt "0"
-  | Dlit f -> Format.fprintf fmt "%f" f
+  | Dlit f -> Format.fprintf fmt "%.19g" f
```
denismerigoux, line 100: Maybe put a comment to explain why you need that here?

denismerigoux (approved):
Ok looks good. Could you say somewhere that even though they compile, dead code elimination and inlining should not be used as is ?

PR #187, opened 2022-09-19
Fix minmax result definition

MdrditArthur:
M functions min & max should return an undefined M value if both arguments are undefined, as it is the behaviour of the old tax system compiler.

### DIFF
--- a/src/mlang/backend_compilers/bir_to_dgfip_c.ml
+++ b/src/mlang/backend_compilers/bir_to_dgfip_c.ml
@@ -341,14 +341,14 @@ let rec generate_c_expr (e : expression Pos.marked)
       let se1 = generate_c_expr e1 var_indexes in
       let se2 = generate_c_expr e2 var_indexes in
       let def_test = Dor (se1.def_test, se2.def_test) in
-      let value_comp = Dfun ("_fmax", [ se1.value_comp; se2.value_comp ]) in
+      let value_comp = Dfun ("max", [ se1.value_comp; se2.value_comp ]) in
       build_transitive_composition
         { def_test; value_comp; subs = se1.subs @ se2.subs }
   | FunctionCall (MinFunc, [ e1; e2 ]) ->
       let se1 = generate_c_expr e1 var_indexes in
       let se2 = generate_c_expr e2 var_indexes in
       let def_test = Dor (se1.def_test, se2.def_test) in
-      let value_comp = Dfun ("_fmin", [ se1.value_comp; se2.value_comp ]) in
+      let value_comp = Dfun ("min", [ se1.value_comp; se2.value_comp ]) in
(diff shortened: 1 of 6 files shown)


src/mlang/backend_compilers/bir_to_dgfip_c.ml
```diff
@@ -819,13 +819,6 @@ let generate_rovs_files (dgfip_flags : Dgfip_options.flags) (program : program)
 #define add_erreur(a,b,c) add_erreur(b,c)
 #endif
 
-#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L)
```
denismerigoux, line 822: Where is the replacement for those?

MdrditArthur: _fmin/_fmax are replaced in the generated code by min/max. 
min/max are already defined as macros (equivalent to the _fmin/_fmax when C < 1999 was used) in the static const.h file, included by the function `generate_header` line 583.

denismerigoux: Ok they're here 

https://github.com/MLanguage/mlang/blob/510b5d457f2a6c42c729c70fdfae7a70e986a60e/examples/dgfip_c/const.h#L109-L110

src/mlang/optimizing_ir/partial_evaluation.ml
```diff
@@ -618,27 +618,13 @@ let rec partially_evaluate_expr (ctx : partial_ev_ctx) (p : Mir.program)
...
-                    (fun a d ->
-                      if Pos.unmark a = Mir.Literal Undefined || d = Undefined
-                      then Pos.same_pos_as (Mir.Literal (Float 0.)) a
-                      else a)
-                    new_args new_ds
-                in
-                let new_e =
-                  Pos.same_pos_as (Mir.FunctionCall (func, new_args)) e
-                in
-                (Pos.unmark new_e, Float)
```
Keryan-dev, line 641: It's a shame not to reduce those when we can.

Please comment that we should do it at some point.

Keryan-dev: Ah, too late. Well then, later I guess.

MdrditArthur: Would you like an issue on the topic ?

Keryan-dev: #194 :)

Keryan-dev: To be clearer, it does not patch multimax. But I don't think it's worth trying to considering how the pass works and how little chance there is to actually simplify it statically.

PR #193, opened 2022-09-23
Smarter C generation

Keryan-dev:
This avoid printing a lot of dumb intermediary assignments by simplifying defineness and computation expression separately.

We could still do more with this approach, such as optimise register allocation but I don't think it's worth its share of work for now.


Keryan-dev:
Alright, I think I'm done with this.

This now have a way to generate intermediary computation much more closely to the legacy compiler, i.e. variable reuse.

A few points:
- types on intermediary variables is not always ideal but safe, I'd argue that we're not helped by the M semantics on that point.
- verifications were not compiled the same way as rules in the legacy compiler, they had a more refined way of handling purely boolean expression. I don't think we should bother for now.
- control is still mostly done in a functional way, this could be improved, but distinguishing between small branches reasonably well-fitted in ternary and more expensive computations is not free.

denismerigoux (approved):
Ok the code changes are readable. What's missing is a set of high-level ocamldoc comments detailing the architecture and philosophy of the DGFiP C backend : how to use the smart constructors, what is a decoupled expr, etc.

Keryan-dev:
Made some docs/comments.
@denismerigoux, you can merge when you're happy.

PR #195, opened 2022-09-26
Fix vanishing not operand

Keryan-dev:
My bad


ddeclerck:
This has been extensively tested and this solves the bugs I encountered. Let's merge !

PR #196, opened 2022-10-13
Fix guarding condition in accessing M table variable

MdrditArthur:
C and Java backend include a condition checking than a M cell table reading is in the table bound. The reference behaviour is to output a M undefined value if the queried cell has an index greater than the table last cell.

The backends where checking for index strictly greater than table size (logically, in practice this or the symmetrical check), as M table are indexed from 0, the last cell has size-1 index and the check should be greater or equal to size.

### DIFF
--- a/examples/java/src/com/mlang/MValue.java
+++ b/examples/java/src/com/mlang/MValue.java
@@ -276,7 +276,7 @@ public class MValue {
     
     if (indexInteger < 0) {
       return zero;
-    } else if (indexInteger > size) {
+    } else if (indexInteger >= size) {
       return mUndefined;
     } else {
       return array[tableStart + indexInteger];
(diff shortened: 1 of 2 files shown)


denismerigoux:
I merged that @MdrditArthur but we should also update this semantics change in the reference BIR interpreter.

MdrditArthur:
Isn't the interpreter already correct?
See bir_interpreter.ml:689 (assignation to the table cell) and bir_interpreter.ml:393 (acces to cell) for index checks, as far as I can tell they are the checks we are talking about and they're already coherent with numbering from 0 : index < 0 and index >= size.

Keryan-dev:
I believe it's already correct in the interpreter. The backends had simply misinterpreted the formal semantics due to a different formulation.

denismerigoux:
Ok perfect!

PR #198, opened 2022-10-28
Use C bindings for TGV in ml code

Keryan-dev:
This will make tests with ml code much faster.

Incidentally, it changes the signature of the TGV manipulation module to have a lot more side effects, which I believe reflects much more the aimed end result.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ test: build
 
 # use: TESTS_DIR=bla make test
 tests: build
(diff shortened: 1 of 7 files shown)


Keryan-dev:
This has been thoroughly tested, with around 5x speed gain on tests execution. This can be merged I think.

denismerigoux (approved):
I see what you did, and why it improves performance. Thanks, let's merge!

PR #194, opened 2022-09-26
Review and fixes on partial evaluation

Keryan-dev:
I spotted some semantic discrepancies on this pass and fixed them, hopefully.

PR #201, opened 2022-11-25
Fix tricky or semantics bug in partial evalutation

Keryan-dev:
This one was a real pain.


ddeclerck:
Alright, this fixes the problem introduced by PR #194. We can merge it.

denismerigoux (approved):
Wouahou I can imagine how painful that was. The validity of these optimizations really ought to be machine checked at some point to prevent things like that.

PR #203, opened 2022-11-28
Update primitive CI, add raw TGV compare

ddeclerck:


ddeclerck:
That's fine by me, shall we merge ?

denismerigoux (approved):
Ok so this adds a debugging feature that compares the end-result TGVs for fuzzign tests. Ok to merge but before the `output.expected` folder should be renamed / put in a sub-dir that tells us which year (2020,2021,2022?) these tests are from.

examples/dgfip_c/ml_primitif/refresh.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+for f in ../../../tests/2020/fuzzing/fuzzer_* ; do
+  if [ ! -d $f ] ; then
+    ./cal $f > /dev/null
+    echo $? $f
+  done
```
mdurero, line 7:
```suggestion
  if [ ! -d $f ] ; then
    ./cal $f > /dev/null
    echo $? $f
  fi
```
Better for my Bash

mdurero, line 7:
```suggestion
  if [ ! -d $f ] ; then
    ./cal $f > /dev/null
    echo $? $f
  fi
```
Better for my Bash

examples/dgfip_c/ml_primitif/promote_tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+year=${1:-2020}
```
ddeclerck, line 3: I guess you could use something like 2000 for lower bound :p

Keryan-dev: It's just the default value, we don't even have M code for 2000.

examples/dgfip_c/ml_primitif/main.ml
```diff
@@ -114,7 +114,7 @@ let run_test test_file =
 
   let annee_calc = M.annee_calc () in
 
-  let out = Printf.sprintf "output/%s.tgv" (Filename.basename test_file) in
+  let out = Printf.sprintf "%d.output/%s.tgv" annee_calc (Filename.basename test_file) in
```
ddeclerck, line 117: (feedback, no change needed) Technically, we can run a calculator compiled for year N on tests for year M where M <> N. I don't know if we want to test that though, in which case it's correct to specify `annee_calc here`. Note that the check below (line 123-124) remains necessary in case a test set for year N contains a test that incorrectly specify a year M <> N.

examples/dgfip_c/ml_primitif/promote_tests.sh
ddeclerck: Alright, I must still be sleeping, I thought it was a range. Seems I've just learnt how to specify default value for arguments in bash.

Keryan-dev:
We need to add an option to the test executable to disable binary comparison in case control results are not available.

denismerigoux (approved):
Ok for me again, merging this finally!

PR #205, opened 2022-12-07
Fix CI failure on updated GH runners

Keryan-dev:
@mdurero found out that some recent changes on github runners would break CI through some stricter clang default options.

The culprit is a single global variable in the C generated code.

### DIFF
--- a/examples/dgfip_c/ml_primitif/build.sh
+++ b/examples/dgfip_c/ml_primitif/build.sh
@@ -86,7 +86,7 @@ fi
 
 echo 'Compilation de la calculette primitive'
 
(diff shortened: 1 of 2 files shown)


denismerigoux (approved):
This looks like someone copy-pasted code from the old compiler but then decided to implement a more proper scheme of local variables instead of the `my_var1` global hack :) But then forgot it and leaved it there.

Keryan-dev:
This was used to the macro definitions for custom float maths like `fabs` in some compiler options. We should keep that in mind in case we should work around that at some point.

mdurero (approved):

PR #206, opened 2022-12-16
Dgfip interoperability

mdurero:
This adds the necessary options to directly use and test Mlang in DGFiP internal context.


denismerigoux (approved):
Merci Mathieu!

PR #207, opened 2022-12-27
Fix generated ROV files path

mdurero:
If Mlang is called with an output filename including a path (say the relative `foo/bar.c`), previously:
- `bar.c` and auxiliary files like `irdata.c`were written in `foo/`
- rules and verification files like `chap-2.c` were written in the working directory of the call.

With this patch, all C source files are written in `foo/`


denismerigoux (approved):
Thanks!

mdurero:
It seemed OK yesterday.
It doesn't work at all today.
Sorry.

PR #209, opened 2023-01-03
Fix failure on java test with preexisting artefacts

Keryan-dev:
Default mkdir does not like feeling useless.


mdurero (approved):
I suppose it was exiting with "make -B". All good.

PR #211, opened 2023-01-13
Fix C type consistency

Keryan-dev:
C generation could lead to some incorrect code if the M input was not type-safe. This patch should handle that properly.

Bonus : less dubious conversions in the code.

### DIFF
--- a/src/mlang/backend_compilers/decoupledExpr.ml
+++ b/src/mlang/backend_compilers/decoupledExpr.ml
@@ -147,8 +147,8 @@ let collapse_constr (st : local_stacks) (ctx : local_vars) (constr : constr) =
 let push_with_kind (st : local_stacks) (ctx : local_vars) (kind : dflag)
     (constr : constr) =
   let expr, ekind, lv = constr st ctx in
-  let st, lv, expr = store_local st lv Anon kind expr in
   let expr = if kind = ekind then expr else cast kind expr in
+  let st, lv, expr = store_local st lv Anon kind expr in
   (st, lv, expr)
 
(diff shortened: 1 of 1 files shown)


denismerigoux:
Ok I'm at the point where I don't understand the code in `decoupledExpr.ml` and I'm unable to even decipher what this patch is doing based on the code changes. Next time I come to Noisy I would be grateful if you could walk me through some of it. For this PR I would actually have expected the backend code to use the type declarations provided in M and stored here:

https://github.com/MLanguage/mlang/blob/02ed2c9db080ef3b983a71292e7e454aa46c6e1f/src/mlang/m_frontend/mast.ml#L352-L358

Keryan-dev:
Of course, next time ! This patch can wait anyway.

About the types, two things :
- the types of `saisie` variables is difficult to propagate to all TGV variables through inference because there is just not enough infos. Even distinguishing boolean from real values might be tricky to do automatically.
- The legacy compiler was forcing consistency over those two types *syntactically* whereas Mlang allows some leeway on this. The problematic case leading to this patch do not happen with the current codebase since it would be rejected in production workflow. We should discuss this at some point.

mdurero:
> I would actually have expected the backend code to use the type declarations provided in M and stored here:

Sadly, a dive in the variable declaration file confirms that type is optional and exceedingly few variables have a type (about 900/35000). 
The associated logic isn't clear to me yet, but of the 900, the full spectrum of variables seems concerned (computed, input, "base" or not, every possible type used at least once…)

denismerigoux:
I had the explanation from @Keryan-dev and I understand now, merging this!

PR #213, opened 2023-02-08
[DGFiP] Allow using -kN with N between 1 and 4 (useful for DGFiP tests) + some cleanup

ddeclerck:
Note: updated expected test results ; they changed because of the removal of calls to `M.sauve_base_XXX` (these are are only used when a corrective computation calls a primitive computation, but we are not in this case here)


examples/dgfip_c/ml_primitif/build.sh
```diff
@@ -87,7 +87,7 @@ else
 
   cd ./calc
 
-  $CCOMPILER -std=c89 -pedantic $CCOPTIONS -O2 -c irdata.c enchain.c var.c contexte.c famille.c revenu.c revcor.c penalite.c variatio.c tableg01.c restitue.c chap-*.c res-ser*.c coc*.c coi*.c horiz*.c
+  $CCOMPILER -std=c89 -pedantic $CCOPTIONS -O2 -c irdata.c enchain.c var.c contexte.c famille.c revenu.c revcor.c penalite.c variatio.c tableg??.c restitue.c chap-*.c res-ser*.c coc*.c coi*.c horiz*.c
```
denismerigoux, line 90: Is the `??` intended or is it bash magic to be replaced by the correct characters later?

examples/dgfip_c/ml_primitif/common.ml
```diff
@@ -266,20 +268,36 @@ module TGV = struct
         set_bool tgv signvar (v < 0.0);
         set tgv dvar (Float.abs v)
 
-  let internal_copy ~ignore_undefined (tgv : t) var_list =
+  let reset_saisie_calc ~except tgv =
+    let save = get_map_opt tgv except in
+    reset_saisie_calculee tgv;
+    set_map tgv save
+
+  type undef_action =
```
denismerigoux, line 276: If I understand correctly this is a sum type specifying the 3 possible intended behavior when we copy a TGV and read from an undefined value, is it not ?

examples/dgfip_c/ml_primitif/stubs.c
```diff
@@ -614,19 +632,98 @@ static void init_var_dict(void)
   sort_index();
 }
 
-void efface_var(T_irdata *irdata, T_var_irdata desc)
+/*
+typedef struct idx_entry_t {
+  short start;
+  short end;
```
denismerigoux, line 638: Why this commented code here?

examples/dgfip_c/ml_primitif/build.sh
ddeclerck: It's just a wildcard character, like `*`. `?` means "any single character".

examples/dgfip_c/ml_primitif/common.ml
ddeclerck: It is. ;)

examples/dgfip_c/ml_primitif/build.sh
mdurero: It's bash magic : one and exactly one character per "?". If k=2, then tableg01.c and tableg02.c are generated, and so on.

examples/dgfip_c/ml_primitif/stubs.c
ddeclerck: It's just experimental code for a poor man's pseudo-hashtable to improve variable lookup times (actually doesn't really change anything in terms of performance). I don't know yet if i'll keep that or not, hence the comment.

examples/dgfip_c/ml_primitif/build.sh
denismerigoux: We never stop learning, thanks :)

ddeclerck: I beg to differ. This is no magic (it would imply we do not understand what happens). It's a feature ;).

denismerigoux:
This needs rebasing but good for me!

PR #216, opened 2023-04-04
[DGFiP] necessary adjustments to build shared object as used by DGFiP API

mdurero:
Part 1 of the tax calculator deliveries to DGFiP clients : tax calculator as a shared object for internal APIs.
Part 2 of the tax calculator deliveries to DGFiP clients : tax calculator as an archive for "modern" mainframes.

Should be mostly changes to the build system to allow more parameters instead of the hard-coded values of the autotest builds.


mdurero:
So, AFAIU this state of the repository is good enough for API and mainframe packages build, providing only a handful of custom Make targets on our side at the start (selected source files) and end (package build from .o). Seems like a good step to merge once clean.

Makefile.include
```diff
@@ -2,7 +2,7 @@
 # Local customization
 ##################################################
 
-SELF_DIR=$(realpath $(dir $(realpath $(lastword $(MAKEFILE_LIST)))))
+SELF_DIR:=$(realpath $(dir $(realpath $(lastword $(MAKEFILE_LIST)))))
```
denismerigoux, line 5: What's the difference in Makefile semantics?

denismerigoux: Shouldn't it be `?=` instead ?

denismerigoux (approved):
Looks good to me!

Makefile.include
mdurero: In Make world, it is a "simply expanded variable". `:=` means the value of `SELF_DIR` is computed (Make says "expanded") once and for all at definition time. Equivalent to say I'm passing the value of the expression as the variable value, and not the expression itself. 
So `SELF_DIR` will stay in the following what it was in this first line.
`=` are recursively expanded. The expression is evaluated when the variable is referenced… so if someone includes a makefile from another directory, then it becomes the last of the `MAKEFILE_LIST` and the `SELF_DIR` is now its directory instead of Mlang's.
I ran into this because my DGFiP config file includes directly one of the old makefile from the existing system, in order to load the C sources file lists corresponding to the different target without duplicating them.

denismerigoux: Wouahou this should be in a comment, otherwise somebody is going to remove the `:` in the future with good intentions but catastrophic consequences :)

mdurero: Example: let's say I'm going `rm -rf` a subdir of Mlang in a clean command and I use `SELF_DIR` as a base of my path. Suddenly I just wiped another totally unrelated and badly-versionned subdir of the server. This is based on real world stunt, do not try to reproduce at home.

mdurero: You can see I used it also in commit [d63a2b2](https://github.com/MLanguage/mlang/pull/216/commits/d63a2b2988503672440229185ff1ebe98571c7fe), in order to ensure shell call is only done once. Otherwise, if you define function in makefiles to call in variable assignments, the function is called every time the variable is used. Not so visible with a short `find` on 30 files, but a bit painful with SVN update or log crawl on the whole repository.

examples/dgfip_c/ml_primitif/Makefile
```diff
@@ -80,12 +78,13 @@ calc/enchain.c: $(SOURCE_FILES)
...
-# Ml_primitif (current main build)
-# -----------------------------
-calc/enchain.o: calc/var_static.c.inc calc/enchain.c
+C_SOURCES_TARGETS?=calc/var_static.c.inc calc/enchain.c
+
+calc/enchain.o: $(C_SOURCES_TARGETS)
 	@echo "Compilation des fichiers C issus des fichiers M (CC=$(CC), CFLAGS=$(CFLAGS))"
 	cd calc && $(CC) $(CFLAGS) -c \
-	irdata.c enchain.c var.c contexte.c famille.c revenu.c revcor.c penalite.c variatio.c tableg??.c restitue.c \
+	*.c
```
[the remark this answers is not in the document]
denismerigoux: No no `*.c` is fine here :)

denismerigoux:
You can merge when you want!

examples/dgfip_c/ml_primitif/Makefile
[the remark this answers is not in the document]
mdurero: Oh, `*.c` was my smallest problem (DGFiP has premade exhaustive source and object file lists from the old calculator build system I can leverage here). I had even forget I was thinking to replace the wildcard by the explicit list…

My question was more on the prerequisite part. I ran in another limitation with mainframe package: we give them C sources, so at the moment I have a useless compilation step in this case. I would say "good enough", justify it by minimally the compilation pass and call it a feature though.

PR #219, opened 2023-05-31
Make possible to rebuild C backend without cleaning the build directory first.

mdurero:
Simple fix:
- calculator C compilation step: build a file list excluding later driver files,
- driver C&ML compilation & linking step: remove object files to be produced by this step at its start.


denismerigoux (approved):
OK!

PR #222, opened 2023-07-05
IRJ files parser as an Mlang sublibrary

mdurero:
Parser for DGFiP test files, AKA "fip files", ".irj", "creneau files", "autotest files"… to load data in a straightforward OCaml record.

Compatible with both actually used file formats right now, primary and corrective computation, and cleaned from unidentified format variations.

Parser is based on original Mlang interpreter parser.

To be used by Mlang interpreter, Mlang OCaml backend, or any external OCaml application.


denismerigoux:
@noeensarguet can you push your changes here so that I can review them?

denismerigoux:
So what you did here is pull all the dependencies of `parser_utils.ml` into `irj_parser` but I don't think we can merge this as it destructures the architecture of Mlang completely. The files : 
* `src/mlang/m_frontend/mast.ml`
* `src/mlang/utils/cli.ml`
* `src/mlang/utils/cli.mli`
* `src/mlang/utils/pos.ml`
* `src/mlang/utils/pos.mli`
Should remain where they are, hence `irj_parser` should not depend on them. Hence you should refactor the code of `parser_utils.ml` to not depend on anything in the list of files above, so that `irj_parser` can really be a standalone library that does not depend on Mlang. For instance, instead of throwing error messages with `Cli.raise_error`, the `irj_parser` library should throw an OCaml exception it defines. It should also embark its mini-AST of the test files.

Keryan-dev:
This is a bit of a tangent.

I agree that depending on `MAst` is too much, and that errors should be handled in a more library-friendly way (i.e. exceptions or result types). `Pos` type and functions could be reused but a different minimal version should be enough for the current purpose.

This aside, I would strongly advocate for a much more librarified architecture for Mlang as we may use some of it to build other helpful tools.

denismerigoux:
I agree with you Keryan, we should move to an Mlang architecture where basically each intermediate representation is its own separate library with clear dependence flow between them. That being said I think this PR is a first step toward this : if when this PR is merged `irj_parser` becomes its own separate library that doesn't depend on the rest of Mlang, we're progressing toward a more modular architecture no?

Keryan-dev:
Yes of course. What I meant is that I would not be shocked to see some adjacent libraries depend on part of the compiler, this one included.

denismerigoux:
Oh yes of course. For instance we'll surely have a `utils` library that all the intermediate representations, and maybe `irj_parser` if we want, could depend on.

noeensarguet:
Should be better. Still quite unsatisfied with the (wrapped false) stanza in the irj_parser library, but this was necessary to open Irj_ast in module Test_interpreter. Also opened the definition of type Pos.t to be able to make it compatible with the type t in Irj_ast

irj_parser.opam
```diff
@@ -1,22 +1,18 @@
 # This file is generated by dune, edit dune-project instead
 opam-version: "2.0"
+version: "1.1.0"
 synopsis: "Parser for the IRJ tests"
 description:
   "This parser is aimed for the tests used by la DGFiP to test the calculation of the French income tax"
-maintainer: ["Mathieu Durero" " Noé Ensarguet"]
```
denismerigoux: Ha ha why did you remove yourself and Mathieu and put me as the maintainer ;)

src/irj_parser/dune
```diff
@@ -0,0 +1,10 @@
+(env
+ (static
+  (ocamlopt_flags
+   (-O3 -ccopt -static))))
```
denismerigoux, line 4: Are we sure we want `-O3` ? That is pretty agressive

src/irj_parser/main.ml
denismerigoux, line 1: Why is there this empty file here?

src/irj_parser/src/dune
```diff
@@ -0,0 +1,15 @@
+(env
+ (static
+  (ocamlopt_flags
+   (-O3 -ccopt -static))))
```
denismerigoux, line 4: Ditto `-O3`

src/irj_parser/src/irj_ast.ml
```diff
@@ -0,0 +1,74 @@
...
+
+   You should have received a copy of the GNU General Public License along with
+   this program. If not, see <https://www.gnu.org/licenses/>. *)
+
+type t = { pos_filename : string; pos_loc : Lexing.position * Lexing.position }
+
+let mk_position sloc =
+  { pos_filename = (fst sloc).Lexing.pos_fname; pos_loc = sloc }
+
+exception
```
denismerigoux, line 22: You don't need to duplicate the full type here as you won't use all its components in `irj_parser`. You can simply declare something like and later do `raise TestParsingError(...)`.

```ocaml
exception TestParsingError of string * pos
```

src/irj_parser/src/irj_ast.ml
```diff
@@ -0,0 +1,74 @@
...
+
+   This program is distributed in the hope that it will be useful, but WITHOUT
+   ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+   FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+   details.
+
+   You should have received a copy of the GNU General Public License along with
+   this program. If not, see <https://www.gnu.org/licenses/>. *)
+
+type t = { pos_filename : string; pos_loc : Lexing.position * Lexing.position }
```
denismerigoux, line 17: That type should not be named `t` as `t`would be the main type of this module, which we expect to be the AST of a whole test file and not a code position.

src/irj_parser/src/irj_file.ml
```diff
@@ -0,0 +1,42 @@
...
+  let filebuf = Lexing.from_channel input in
+  let filebuf =
+    {
+      filebuf with
+      lex_curr_p = { filebuf.lex_curr_p with pos_fname = test_name };
+    }
+  in
+  let f =
+    try Irj_parser.irj_file Irj_lexer.token filebuf with
+    | StructuredError e ->
```
denismerigoux, line 30: Here you can catch `TestParsingError` raised in the parser

src/irj_parser/src/irj_file.ml
```diff
@@ -0,0 +1,42 @@
...
+        close_in input;
+        raise (StructuredError e)
+    | Irj_parser.Error ->
+        close_in input;
+        raise
+          (StructuredError
```
denismerigoux, line 36: And here raise a `TestParsingError`

src/irj_parser/src/irj_parser.mly
```diff
@@ -0,0 +1,145 @@
...
+   FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+   details.
+
+   You should have received a copy of the GNU General Public License along with
+   this program. If not, see <https://www.gnu.org/licenses/>. *)
+
+%{ open Irj_ast
+
+  let error (sp, ep) msg =
+    raise (StructuredError ("Parse error : " ^ msg, [ (None, mk_position (sp, ep)) ], None))
```
denismerigoux, line 21: Here you can raise a `TestParsingError ...`

src/mlang/test_framework/test_interpreter.ml
```diff
@@ -13,31 +13,15 @@
...
-        close_in input;
-        raise (Errors.StructuredError e)
-    | Test_parser.Error ->
-        close_in input;
-        Errors.raise_spanned_error "Test syntax error"
-          (Parse_utils.mk_position (filebuf.lex_start_p, filebuf.lex_curr_p))
-  in
-  close_in input;
-  f
+let convert_pos (sloc : Irj_ast.t) : Pos.t =
```
denismerigoux, line 18: Yes perfect!

src/mlang/utils/pos.mli
```diff
@@ -18,7 +18,9 @@
 
 (** {2 Source code position} *)
 
-type t
+type t = { pos_filename : string; pos_loc : Lexing.position * Lexing.position }
+(* opened this type. Not very satisfactory but necessary to enforce the
```
denismerigoux, line 22: Instead of opening the type you can just use the `make_position` function below in `test_interpreter.ml/convert_pos` :)

denismerigoux (commented):
Hi @noeensarguet, thanks very much for the corrections! We're getting there, but there are a few cleaning items that should be done before merging this. This is mostly nitpicking but let's try to make the code as beautiful as possible to set the tone for your beginnings on Mlang :)

Keryan-dev:
I don't think you should need a (wrapped false) to be able to use those module elsewhere. Without it you'd have an additional indirection through the top module `Irj_parser` : `open Irj_parser.Irj_ast` didn't do the trick ?

src/irj_parser/dune
Keryan-dev: This stanza is there in a lot of places and only affect static profile builds (only uses for releases IIRC).

irj_parser.opam
noeensarguet: Hahaha
I had a separate dune-project for irj_parser in the first commit, until I realised (thanks to @Keryan-dev) that you could declare both packages in the same dune-project file (and so with the same maintainers etc.)

src/irj_parser/main.ml
noeensarguet: Thought we might want an executable at first and didn't delete it afterwards ^^'

noeensarguet:
@Keryan-dev tried what you said and a couple other things to address the (wrapped false) - maybe I didn't quite wrap my head around this. If I take the stanza out, the only module I can open is Irj_file since it's the top module of the library Irj_parser (and open Irj_parser yields an Unbound module error). Neither open Irj_parser.Irj_ast or open Irj_file.Irj_ast are recognised as bound modules

Besides, if it is possible to add an indirection with open Irj_parser.Irj_ast, I don't quite see what the purpose of the stanza (wrapped _) would be in the first place - so if adding some indirection is an option, I would be down to have some clarifications about that :)

src/irj_parser/dune
noeensarguet: Just checking, are we leaving this as is? I changed the rest following your remarks (thanks for those!) and it's ready to be pushed

denismerigoux: Yes you can leave this as is :) Please push!

PR #223, opened 2023-07-06
Re-inline division to avoid executing division by zero in dgfip backend

Keryan-dev:
Patching the stack allocation mechanism to push down divisions back in their guards.
Because apparently some hardware choose to interrupt on such float division instead of returning infinity.


mdurero:
It passed the test to build and execute on z/OS Unix platform, so I think it’s all good.

PR #224, opened 2023-09-01
WIP: Add a test file checker

ddeclerck:
TODO
- [ ] Check characters out of desired encoding
- [ ] More checks on validity of fields
- [ ] Check name length (has a maximal length in practice)

### DIFF
--- a/src/mlang/test_framework/test_lexer.mll
+++ b/src/mlang/test_framework/test_lexer.mll
@@ -19,50 +19,90 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 {
 open Lexing
 open Test_parser
+
+module StrMap = Map.Make (String)
+
+let keywords =
+  List.fold_left (fun map (kw, tok) ->
+    StrMap.add kw tok map
+  ) StrMap.empty [
+    "#NOM",                 NOM;
+    "#FIP",                 FIP;
+    "#ENTREES-PRIMITIF",    ENTREESP;
+    "#CONTROLES-PRIMITIF",  CONTROLESP;
+    "#RESULTATS-PRIMITIF",  RESULTATSP;
+    "#ENTREES-CORRECTIF",   ENTREESC;
+    "#CONTROLES-CORRECTIF", CONTROLESC;
+    "#RESULTATS-CORRECTIF", RESULTATSC;
+    "#ENTREES-RAPPELS",     ENTREESR;
+    "#CONTROLES-RAPPELS",   CONTROLESR;
+    "#RESULTATS-RAPPELS",   RESULTATSR;
+    "#DATES",               DATES;
+    "#AVIS_IR",             AVISIR;
+    "#AVIS_CSG",            AVISCSG;
+  ]
+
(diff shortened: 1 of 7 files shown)


ddeclerck:
@mdurero Any checks to add ?

mdurero:
I’m finally able to look at this. I’ll try to double check with workmates for what kind of mistake we can ward against.

src/mlang/test_framework/test_lexer.mll
```diff
@@ -19,50 +19,90 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
...
+
+| '*' any* nl
+  { check_cr lexbuf; new_line lexbuf;
+    if is_bol lexbuf then token lexbuf
+    else error lexbuf "Comment with * must start in the first column" }
+
+| blank any* nl
+  { check_cr lexbuf; new_line lexbuf;
+    if is_bol lexbuf then error lexbuf "Line can not start with a blank"
+    else NL }
```
mdurero, line 78: The file specification says:

> Il est possible d'insérer des commentaires dans le fichier de deux façons :
>
>  Ligne de commentaire: faire précéder le commentaire du signe '*'
>  Commentaire en fin de ligne : faire précéder le commentaire d'un espace.

meaning we would have allowed end-of-line comment, starting with space.

I do not find this particularly smart, but then, if we enforce 1 token or record per line, space is possible as a marker.

If I understand correctly that’s exactly what you have done here.

mdurero:
Ideas from a run in comparison to the faulty files we had seen last year:

- when a possibly empty field in a "rappel" is missing, basically a slash was forgotten, perhaps we could detect it (now it's just a generic parser error). Basically could we detect if a "rappel" has one slash more or less than needed?
- variable names are SYMBOL at the moment. I would think they are more restricted, so we could make a specific token for them. Then if we have a SYMBOL (or whatever not being a variable name) which is not a header keyword following a variable/value we can throw an error like « expecting header ». It kind of catches the missing #, or full header line instead of letting them be generic.
- a lot more difficult: date mismatch in "rappel". Simple when event number is a date, otherwise rely on external knowledge of the targeted year. We could propose to enforce ANREV.
- do we enforce ASCII? Seems the case now but in NAME field, perhaps we could allow ISO8859. In any case I’m interested in improving the "Unexpected character" error message to actually display the character hexadecimal code point, that’s easier to read.
- : character could be allowed in name. Number also. Actually "455" could be a valid name.
- a single # character, like #ND, becomes double in error message, like `Unknown section name: '##ND'`

mdurero:
If we take into account the error test files, "ano" folder, a lot of them are just incomplete… they don’t use the specified format. And not only as a way to check the program behaviour. Basically, the question is to consider some of the file parts optional or not. I would prefer not to.

denismerigoux:
Closing in favor of #224.

mdurero:
Just to be clear: this PR's content is not entirely merged (nor should it be, IMHO) in #222, as the file checker module itself is not included in the parser.

PR #228, opened 2024-01-09
Comparaison des flottants à epsilon près (interpréteur + backend C).

mdurero:
Work from @david-michel1, for mainframe interoperability and future maintainers mental sanity.

Must be linked to the corresponding merge request on the private repository with static DGFiP C interface.

### DIFF
--- a/src/mlang/backend_compilers/decoupledExpr.ml
+++ b/src/mlang/backend_compilers/decoupledExpr.ml
@@ -308,7 +308,7 @@ let comp op (e1 : constr) (e2 : constr) (stacks : local_stacks)
         if String.equal op "==" && v1 = v2 then Dtrue else Dbinop (op, e1, e2)
     | _ -> Dbinop (op, e1, e2)
   in
-  let e =
+  let _e =
     match op with
     | "==" -> comp ( = )
     | "!=" -> comp ( <> )
@@ -318,7 +318,7 @@ let comp op (e1 : constr) (e2 : constr) (stacks : local_stacks)
     | ">" -> comp ( > )
     | _ -> assert false
   in
-  (e, Def, lv2 @ lv1)
+  (Dbinop (op, e1, e2), Def, lv2 @ lv1)
 
 let dfun (f : string) (args : constr list) (stacks : local_stacks)
     (ctx : local_vars) : t =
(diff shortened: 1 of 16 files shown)


src/mlang/backend_compilers/decoupledExpr.ml
```diff
@@ -308,7 +308,7 @@ let comp op (e1 : constr) (e2 : constr) (stacks : local_stacks)
         if String.equal op "==" && v1 = v2 then Dtrue else Dbinop (op, e1, e2)
     | _ -> Dbinop (op, e1, e2)
   in
-  let e =
+  let _e =
```
denismerigoux, line 311: The definition of `_e` should be deleted since it's no longer useful.

src/mlang/backend_ir/bir_roundops.ml
```diff
@@ -14,6 +14,8 @@
    You should have received a copy of the GNU General Public License along with
    this program. If not, see <https://www.gnu.org/licenses/>. *)
 
+let epsilon = 0.000001
```
denismerigoux, line 17: So this value should be cutsomizable, and we can piggyback on the existing `--test_error_margin` CLI parameter here : https://github.com/MLanguage/mlang/blob/master/src/mlang/utils/cli.ml#L174

denismerigoux (approved):
It's way saner than before, thanks @david-michel1 !

src/mlang/backend_compilers/decoupledExpr.ml
Keryan-dev: And so is `comp` above. This removes the few optimisations made in this constructor, although I'm not sure it actually triggers on the codebase anyways.

mdurero: If we want to stay as close as possible to the old generator, perhaps keeping the optimisation would be better.

Keryan-dev: The new old, I don't think compir has any such optimisation on comparison operators.
This remains to be checked but the optimisations relied on strict equality, so I think they should not interfere with an epsilon approach (at worst the optimisation doesn't trigger although it always will at runtime thanks to the epsilon).

mdurero: Ok. It seems Denis just chose to use the interpreter's epsilon comparison anyway.

denismerigoux:
Careful : this `epsilon` parameter affects directly the operational semantics of the language, so it should be consistent between the interpreter and the backend. Hence, Mlang should output for the DGFiP a C header file that defines the `EPSILON` value according to how it was called during compilation. @ddeclerck since you reviewed the other PR on `brouillon-correctif` could you add on this PR a feature to do this so that it would play out with the static files on `brouillon-correctif` ?

denismerigoux:
I guess this header file could be generated from here: https://github.com/MLanguage/mlang/blob/8e98f0adf777be3003aba6845ae7cd5c8861816d/src/mlang/backend_compilers/dgfip_gen_files.ml#L1256

ddeclerck:
> Careful : this `epsilon` parameter affects directly the operational semantics of the language, so it should be consistent between the interpreter and the backend. Hence, Mlang should output for the DGFiP a C header file that defines the `EPSILON` value according to how it was called during compilation. @ddeclerck since you reviewed the other PR on `brouillon-correctif` could you add on this PR a feature to do this so that it would play out with the static files on `brouillon-correctif` ?

I guess I could add that in the generation of "config.h", since that file is (should be ?) included by all the static files.

denismerigoux:
Ok I've finally found the time to add the generation of `EPSILON` by `conf.h`. We can now merge this as well as https://github.com/MLanguage/brouillon-correctif/pull/5

PR #233, opened 2024-03-26
CFIR flag does not exclude Iliad flag if from the application option.…

### LINKED ISSUE #231: [DGFiP] Flags for Iliad and CFIR applications are not mutually exclusive in all cases
https://github.com/MLanguage/mlang/blob/46af3c69a6557f5a09089ea2fb3233c4d0ca1895/src/mlang/utils/dgfip_options.ml#L189-L191

In modern (post 2017 I think) builds targeting CFIR-class clients, the old system put both flags to true, due to option `-Ailiad` taken into account after the `-U`, `-R` and other flag-type options. I’m not convinced this is not a bug, but it’s there.

A probably better condition:
``` C
flg_iliad = 
   (iliad_pro && (not cfir))
   || application_name = "iliad"
```
Batch should stay as it is excluding Iliad.

mdurero:
… Fix #231

To  be applied also on mill-2023, either ASAP if deemed safe, or after the batch tax version is freezed.

### DIFF
--- a/src/mlang/utils/dgfip_options.ml
+++ b/src/mlang/utils/dgfip_options.ml
@@ -187,8 +187,7 @@ let handler (income_year : int) (application_name : string) (iliad_pro : bool)
     annee_revenu = income_year;
     flg_correctif = not primitive_only;
     flg_iliad =
-      ((iliad_pro && (not cfir))
-      || application_name = "iliad" )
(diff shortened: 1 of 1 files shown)


denismerigoux (approved):
All good for me!

PR #234, opened 2024-04-12
Suppression fichiers statiques, ajout submodule

noeensarguet:


denismerigoux (approved):
OK for me, thanks!

noeensarguet:
We should also remove the script `UpdateMlangDeps.sh` (handled by the submodule now). Will address this when David comes back

PR #240, opened 2024-06-13
 Refactor IRJ parser for Menhir new error handling

### LINKED ISSUE #238: Menhir "error" token is deprecated
The traditional error handling way of Menhir (Ocamlyacc-style) is deprecated.

The IRJ file parser should be rewritten (again, and again, and again…) using incremental API and .messages file, as per [Error handling: the new way](http://gallium.inria.fr/~fpottier/menhir/manual.html#sec:errors:new).

As far as I understand, the source file parser doesn’t use any of the Menhir error systems.

mdurero:
As stated in #238, traditionnal error handling is deprecated in favor of a flexible approach using Menhir incremental API.

Close #238


mdurero:
The current architecture is kind of underwhelming. 

On the one hand I would find satisfying to merge the `parse_file` and `dummy_parse_file_with_incremental` in a unique parser entry point, to reuse the lexbuf for instance.
On the other hand, in Mlang interpreter context and other "not validation" ones, do we want the full two pass parser to be called when the first pass would be enough ?

Yes, writing it I just realise that the second pass is called only on errors, so it doesn’t slow the interpreter in a right case. But I keep the question as my problem is more than both pass have a very distinct behaviour. The second pass prints on error output in all cases.

So at the end, my proposition would be:
- modify second pass to create a proper error type with its message, but do nothing with it,
- the irj_checker displays all errors,
- other context like the interpreter can choose to not do that and fail silently, or be less verbose.

At the end, I feel like irj_checker could also almost be just a wrapper calling `parse_file` with a verbose flag…

mdurero:
I think we should slow the work on the json generator and focus on the polish on file validation to merge it soon.

denismerigoux (approved):
OK, let's merge!

PR #241, opened 2024-06-28
Parser messages and polish

mdurero:
Now we have it, make it intelligible for end users writing the tests.


denismerigoux (approved):
Hope this message-filling morning was pleasant :)

PR #251, opened 2025-01-17
Update ocaml-setup version

noeensarguet:


noeensarguet:
Ideally, take the time to find a way to *not* run `git submodule update` when setting up opam 2.3. Working fix for now

PR #252, opened 2025-01-27
New errors and warnings

noeensarguet:
- Raise a warning if a variable is defined twice within the same rule
- Raise an error if a variable is defined in two different rules (within the same chaining). This brings back the same behaviour as the legacy compiler.
- Improve the error message for an auto-cycle (the warning typically happens if instructions within a rule are in the wrong order, so that a variable is used before being defined)


david-michel1 (approved):
OK

PR #263, opened 2025-04-29
Fix interpretation of undefined variables

Stevendeo:
Fuzzer tests before 2020 used 0.000000 as an expected value for 'undefined'. Since https://github.com/MLanguage/mlang/pull/262 this equivalence have been lost and make tests fails.

This PR adds more details on variable status after intepretation and accepts 'Undefined' as a correct value for 0.


noeensarguet:
Thanks! Added a couple commits to make the interpreter raise an actual error when running on multiple tests (see https://github.com/MLanguage/mlang/actions/runs/14727382198/job/41333215788 that floods you with error messages on the interpreter run but still has return status 0 :sweat_smile:)

Stevendeo:
Can we merge it?

noeensarguet:
Yes, sorry. Thanks for the follow-up

PR #265, opened 2025-06-06
[WIP] allow use of ocamlformat 0.20

lambdalex-ocp:
# Rationale

On newest version of the libc, the ocaml compiler version 4.13 is broken. A fix
is present in ocaml 4.14.2.

As of now, the only thing preventing the installation of 4.14.2 is the dependency
on ocamlformat 0.19.0. This PR proposes to allow ocamlformat version 0.20, which
would allow the usage of ocaml 4.14.2. Of course, we aim to be fully retrocompatible,
which means that current users of ocamlformat 0.19.0 and ocaml 4.13 should notice
no difference.


Stevendeo:
The simpler would be to upgrade to a version that at least allow to use any version of ocaml 4.xx. The 0.22.4 seems to be the smallest one with that property : https://opam.ocaml.org/packages/ocamlformat/ocamlformat.0.22.4/

Stevendeo (approved):

mdurero:
I see no reason, once we upgrade ocamlformat, to not upgrade as far as we can. Is there an argument against ocamlformat 0.27? The minimum compiler version is the same as 0.22.4.

lambdalex-ocp:
I've updated it to the latest ocamlformat version, `0.27.0`

lambdalex-ocp:
Actually, ocamlformat.0.25.1 fixes some stuff, which would reformat some of our code:

Nested non-indented ifs such as the one found in `dgfip_compir_files.ml` would be indented.

```ocaml
  (if flags.Dgfip_options.flg_debug then
   if flags.nb_debug_c <= 0 then
     let nb = match nb_debug with [ nb ] -> nb | _ -> assert false in
     Format.fprintf fmt "#define NB_DEBUG %d\n" nb
   else
     let i =
       List.fold_left
         (fun i nb ->
           Format.fprintf fmt "#define NB_DEBUG%02d %d\n" i nb;
           i + 1)
         1 nb_debug
     in
     assert (i = flags.nb_debug_c + 1));
     [...]
```

As such, I've bumped ocamlformat to the highest version which causes no breaking changes, 0.24.1.

PR #266, opened 2025-06-10
[WIP] Add a CI

lambdalex-ocp:
This MR adds a CI (Continuous Integration) to the project. As such the test suite is ran whenever a merge request is opened or updated.

This MR is based on the #265 MR, so it should not be merged yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,31 @@
+name: Ocaml CI
+
+on:
+  - pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout tree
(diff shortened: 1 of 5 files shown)


lambdalex-ocp:
You're right, for some reason I didn't find the already present Github action.

We should move away from Github? Where to?

mdurero:
> We should move away from Github? Where to?

Well, let’s say we could wish for philosophical reasons.

Where… the one and only place could be https://gitlab.adullact.net/dgfip/, as a forge both publicly readable on the internet and managed by French administration-adjacent people that we already use.
Now I also know about https://gitlab.mim-libre.fr/explore

mdurero:
May I suggest we close the PR?
Or do you want to use it for workflow improvements?

PR #270, opened 2025-07-16
No local var option

Stevendeo:
This PR adds a new `--no-local-var` option that forces mlang not to define local vars when calculating expressions.
Expressions were compiled as sets of simple binary operations of the form `inti = intj || int k;`. While this is memory efficient, I believe having the whole expression could lead to some optimizations. 

* C compilers can optimize code with smarter heuristics. For example, when checking the definition of a given variable, the current code calculates all the possible branches while some of them could be simplified quickly (if `DC_[A] = 1`, then `DC_[A] || DC_[B] || DC_[C] || DC_[D]` can be evaluated to `1` directly, while `int0 = DC_[A] || DC_[B]; int1 = int0 || DC_[C]; int0 = int1 || DC_[D];` requires to calculate everything.) 

* We could simplify boolean expressions with some basic boolean algebra.

Ideally, the `--no-local-var` option is meant to be temporary. Either the optimization makes the code more efficient and the option becomes the default behavior, or this PR will document why mlang behaves like this.


Stevendeo:
Résultats des tests via `make test_dgfip_c_backend` en comparant les temps d'exécution des tests (100 itérations), avec clang v15.0.7 et gcc 13.3.0 (le pourcentage indique l'acceleration que permet l'optimisation, ou la deceleration s'il est négatif) : 

* clang -O1 : +1.75%
* clang -O2 : -0.84% 
* clang -O3 : +6.3%
* gcc -O1 : +0.98%
* gcc -O2 : +6.26%
* gcc -O3 : +4.92%

Stevendeo:
@david-michel1 La PR a eté rebasée sur adaptation_correctif pour gagner du temps sur le merge, tu peux la review et la merger quand tu veux !

PR #275, opened 2025-08-26
[documentation] [pos.mli]

lambdalex-ocp:
Hello, I've started some work on documenting some files in the codebase. Here is the first of them, `pos.mli`.

### DIFF
--- a/src/mlang/utils/pos.mli
+++ b/src/mlang/utils/pos.mli
@@ -13,60 +13,108 @@
    License for the specific language governing permissions and limitations under
    the License. *)
 
-(** A position in the source code is a file, as well as begin and end location
-    of the form col:line *)
+(** {1 Source Code Position}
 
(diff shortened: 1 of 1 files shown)


src/mlang/utils/pos.mli
```diff
@@ -13,60 +13,108 @@
...
+(** [retrieve_loc_text pos] reads the source file associated with [pos] and
+    returns a formatted string of the code at that location, with the exact
+    columns highlighted. This is used to display code snippets in error
+    messages. *)
+
+(** {2 Marked Value Manipulators} *)
 
 val none : t
-(** Placeholder t *)
+(** [none t] creates a placeholder postion *)
```
Stevendeo, line 67: It does not create it per-se, it *is* the null position.

Stevendeo (approved):
Except for the `none` rewording, looks good!

